PR #112, opened 2014-06-23
Fix ephemeral mounts

zeridon:
Add all possible ephemeral block mounts during AMI registration

Value for this being if not using directly AWS to launch a server (ergo skipping the assignment).

Prime case is Rightscale: they create their images with 1-4 ephemeral mounts. Due to this on some instance sizes you don't see all the available ephemeral storage and there is no way to add it afterwards.

### DIFF
--- a/bootstrapvz/providers/ec2/manifest-schema.json
+++ b/bootstrapvz/providers/ec2/manifest-schema.json
@@ -9,6 +9,9 @@
 			"properties": {
 				"description": {
 					"type": "string"
+				},
+			"ephemerals": {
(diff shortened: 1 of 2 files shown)


zeridon:
Unfortunately i have to retract the pull

andsens:
How so? What changed?

udangel-r7:
any idea why this was removed? i am looking for something similar atm

andsens:
@udangel-r7 we're talking more than 2 years ago man, no clue :-/

PR #162, opened 2014-11-22
Add support for enhanced networking on EC2

impl:
Hi all,

I've added support as a plugin for EC2's enhanced networking using the new Intel driver. The steps in the plugin follow the [official EC2 documentation](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/enhanced-networking.html) for enabling it.

The existing enhanced networking installation task wasn't used (and didn't work), so I've also removed that.

On my test instances, enhanced networking reduces latency by 10-20% and jitter by almost 90%.

This feels like something that should be automatically enabled for EC2/HVM configurations, so I'd be happy to refactor it to the EC2 provider as well. I believe Ubuntu enables it by default for HVM.

Thanks!

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,6 @@
+2014-11-23:
+	Noah Fontes:
+		* Add support for enhanced networking on EC2 images
 2014-07-12:
 	Tiago Ilieve:
 		* Fixes #96: AddBackports is now a common task
--- a/bootstrapvz/common/tasks/kernel.py
+++ b/bootstrapvz/common/tasks/kernel.py
@@ -1,0 +1,26 @@
+from bootstrapvz.base import Task
+from .. import phases
+from ..tasks import packages
+
+
+class AddDKMSPackages(Task):
+	description = 'Adding DKMS and kernel header packages'
+	phase = phases.package_installation
+	successors = [packages.InstallPackages]
+
+	@classmethod
+	def run(cls, info):
+		info.packages.add('dkms')
+		kernel_pkg_arch = {'i386': '686-pae', 'amd64': 'amd64'}[info.manifest.system['architecture']]
+		info.packages.add('linux-headers-' + kernel_pkg_arch)
+
+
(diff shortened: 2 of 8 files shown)


daveloyall:
So, did you end up using the host's build environment for this?  I think so, because today (using HEAD of master) I got this:

    Error! Your kernel headers for kernel 3.2.0-4-amd64 cannot be found.
    Please install the linux-headers-3.2.0-4-amd64 package,
    or use the --kernelsourcedir option to tell DKMS where it's located
    Command 'chroot /target/c670114b/root dkms build -m ixgbevf -v 2.16.1' returned non-zero exit status 1

The host is:

    hobbes@somemachine:~/bootstrap-vz$ uname -a
    Linux someothermachine 3.2.0-4-amd64 #1 SMP Debian 3.2.41-2 x86_64 GNU/Linux

The guest (target new AMI) is (this is from the bootstrap-vz's --debug console output):

    ...    
    Determining kernel version
    Kernel version is 4.0.0-2-amd64
    ...

Wait a minute...  Here's the host again:

    hobbes@somemachine:~/bootstrap-vz$ uname -v
    #1 SMP Debian 3.2.41-2

Where does `${kernelver}` come from?  https://github.com/andsens/bootstrap-vz/blob/master/bootstrapvz/providers/ec2/tasks/network.py#L56

To be clear: it appears that dkms is attempting to target a kernel version that is neither that of the target nor that of the host.  (Or, I'm reading this wrong! :) )

daveloyall:
`${kernelver}` in `network.py` might be a red herring?  I manually altered it (hardcoded it to the target's kernel version) but nothing changed.

This change to my manifest seems to have done the trick for me:

    provider:
      enhanced_networking: none

I expect that I can install the fancy network drivers at runtime.

andsens:
> I expect that I can install the fancy network drivers at runtime.

You can, though this should work. I created #240 to track the progress on this issue.

Issue #127, opened 2014-07-29
parted refuse to create the last partition


vincentbernat:
Hi!

Using `kvm-virtio.manifest.yml`, building an image fails with the following output:

```
Creating a loopback volume
Executing: qemu-img create -f raw /target/39b141ce/volume.raw 1024M
Formatting '/target/39b141ce/volume.raw', fmt=raw size=1073741824
Attaching the volume
Executing: losetup --show --find /target/39b141ce/volume.raw
/dev/loop0
Partitioning the volume
Executing: parted --script --align none /dev/loop0 -- mklabel msdos
Executing: parted --script --align none /dev/loop0 -- mkpart primary 2MiB 32MiB
Executing: parted --script /dev/loop0 -- set 1 boot on
Executing: parted --script --align none /dev/loop0 -- mkpart primary 32MiB 160MiB
Executing: parted --script --align none /dev/loop0 -- mkpart primary 160MiB 1GiB
Error: The location 1GiB is outside of the device /dev/loop0.
Command 'parted --script --align none /dev/loop0 -- mkpart primary 160MiB 1GiB' returned non-zero exit status 1
Traceback (most recent call last):
  File "/home/bernat/code/exoscale/bootstrap-vz/bootstrapvz/base/main.py", line 96, in run
    tasklist.run(info=bootstrap_info, dry_run=dry_run)
  File "/home/bernat/code/exoscale/bootstrap-vz/bootstrapvz/base/tasklist.py", line 38, in run
    task.run(info)
  File "/home/bernat/code/exoscale/bootstrap-vz/bootstrapvz/common/tasks/partitioning.py", line 27, in run
    info.volume.partition_map.create(info.volume)
  File "/home/bernat/code/exoscale/bootstrap-vz/bootstrapvz/base/fs/partitionmaps/abstract.py", line 50, in create
    self.fsm.create(volume=volume)
  File "/usr/lib/python2.7/dist-packages/fysom/__init__.py", line 178, in fn
    if self._before_event(e) is False:
  File "/usr/lib/python2.7/dist-packages/fysom/__init__.py", line 208, in _before_event
    return getattr(self, fnname)(e)
  File "/home/bernat/code/exoscale/bootstrap-vz/bootstrapvz/base/fs/partitionmaps/msdos.py", line 64, in _before_create
    partition.create(volume)
  File "/home/bernat/code/exoscale/bootstrap-vz/bootstrapvz/base/fs/partitions/base.py", line 43, in create
[...]

PR #172, opened 2014-12-11
Fix an off-by-one error with computation of partition end.

andrewbogott:
Alternative fix for bug #127

### DIFF
--- a/bootstrapvz/common/fs/loopbackvolume.py
+++ b/bootstrapvz/common/fs/loopbackvolume.py
@@ -11,7 +11,7 @@ class LoopbackVolume(Volume):
 
 	def _before_create(self, e):
 		self.image_path = e.image_path
-		vol_size = str(self.size.get_qty_in('MiB')) + 'M'
+		vol_size = str(self.size.get_qty_in('MiB') + 1) + 'M'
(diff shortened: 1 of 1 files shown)


EmmanuelKasper:
As an alternative to enlarging the qemu disk image, we could also the parted special adress -0 to adress the end of the disk

This fails on a 1GB disk image using parted 3.2

    parted --script --align none /dev/loop0 -- mkpart primary 160MiB 1GiB
    Error: The location 1GiB is outside of the device /dev/loop0.

but this works

    parted --script --align none /dev/loop0 -- mkpart primary 160MiB -0

The problem of usinf the  `-0` option we would need to change the call to _before_create  in partitions/base.py to handle the case of bering the last partition of the disk image.

So after a bit of thought I think your solution is better.

andsens:
[Check this out](https://github.com/andsens/bootstrap-vz/blob/wip-integration-testing/bootstrapvz/base/fs/partitionmaps/gpt.py#L65). It still needs a little work before I'm ready to merge, but the parted error is fixed by simply subtracting a single section from the last partition.

andsens:
Fixed in `wip-integration-testing`.

PR #188, opened 2015-01-27
Recover file permissions after shrink a vmdk image

jwendell:
After vmware-vdiskmanager runs, the image is left with a 600
permission. This fixes it by applying a 644 chmod on it.


andsens:
Hm, I think it would be much better to *preserve* the permissions, i.e. read them before shrinking and then reapply them. This would mean any future code that changes permissions on the image won't get broken by this fix.

Also: You need to send the PR to the `development` branch.

PR #190, opened 2015-01-27
Recover file permissions after shrink a vmdk image

jwendell:
After vmware-vdiskmanager runs, the image is left with a 600
permission. This commit fixes it by preserving the file permission
before the shrink operation.


andsens:
Hah, that was quick :-)  
Thank you for the contribution! Merging...

PR #192, opened 2015-01-30
Disable resize on disks larger than 2TB

wrigri:
This is a temporary fix to work around a bug in Debian where growpart can destroy a disk >2TB since it updates the partition table incorrectly (i.e. it can shrink the partition under some circumstances.

This bug is described in more detail in this thread: https://lists.debian.org/debian-cloud/2014/10/msg00018.html

Once a fix has been submitted to Debian, this change can be reverted.

Quick summary of the change:
This change adds a file to initramfs in /etc/initramfs-tools/scripts/local-premount/ called gce-disable-growroot. This is a shell script that checks the size of the disk and if it's greater than 2TB, it touches the file /etc/disable-growroot on the disk. When growroot discovers that file, it will not attempt to resize the root partition.

### DIFF
--- a/bootstrapvz/providers/gce/__init__.py
+++ b/bootstrapvz/providers/gce/__init__.py
@@ -9,6 +9,7 @@ import tasks.packages
 from bootstrapvz.common.tasks import apt
 from bootstrapvz.common.tasks import loopback
 from bootstrapvz.common.tasks import initd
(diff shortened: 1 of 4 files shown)


andsens:
On vacation right now, but I trust you guys at google have somewhat tested this :-)  
Merging...

PR #193, opened 2015-02-05
Don't boot quietly on extlinux (aws)

jwendell:
So that we are able to see the System Log through EC2 console
or command line tools.


andsens:
Merging... I'm doing something along the same lines on the integration-test branch, but I'll just fix whatever conflict there is when I merge that.

PR #194, opened 2015-02-11
Added an Ansible plugin, which runs a playbook on the chroot

Rory-Finnegan:
I'll update the dev_Plugins section with the following info once this gets merged.

## Ansible ##

This plugin runs the supplied ansible playbook in the chrooted environment similar to the puppet plugin.

### Setting ###
  * `playbook`: the path to the playbook on the host OS. ex) `"/opt/playbooks/webserver.yml"`
    * required
  * `hosts`: If you have your playbook broken up into different plays, specify the hostnames of the plays you want to run on the resulting image. ex) `['webserver']`
  * `extra_vars`: You can supply extra vars to the playbook here with the same syntax. ex) `'foo=bar'`
  * `tags`: Limit what runs on the playbook by providing the tags for tasks you'd like to run as a comma separated string. ex) `'install,configure'`
  * `skip_tags`: Same as `tags`, but excludes those tasks.
  * `opt_flags`: While not recommended, you can supply any other flags to the ansible playbook as an array of strings.  ex) `['-vvvv']`

### DIFF
--- a/bootstrapvz/plugins/ansible/__init__.py
+++ b/bootstrapvz/plugins/ansible/__init__.py
@@ -1,0 +1,13 @@
+import tasks
+
+
+def validate_manifest(data, validator, error):
+    import os.path
+    schema_path = os.path.normpath(os.path.join(os.path.dirname(__file__), 'manifest-schema.yml'))
+    validator(data, schema_path)
+
+
+def resolve_tasks(taskset, manifest):
+    taskset.add(tasks.AddPackages)
+    taskset.add(tasks.CheckPlaybookPath)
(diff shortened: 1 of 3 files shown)


Rory-Finnegan:
Found an issue with extra_vars in the chroot.  The other flags are working fine and the command created works fine if I run it outside the chroot, but for some reason the extra_vars in the command are undefined in the chroot.

andsens:
This is really cool, I can't wait to get some preprovisioned images up and running at work :-)
Merging...

p.s.: The build failing was an issue with the existing code, I think PEP8 changed a bit.

PR #198, opened 2015-03-03
Add the manifest "include-source-type" key for packages object

jwendell:
It controls whether to include the 'deb-src' lines in image's
source.list.

Currently they are always included. This patch changes this
behavior by not including them by default; the user must set
this new config to true in order to include them.

This saves a bit of bandwidth in default installations. Also,
the use of src packages is not so usual in ordinary installations.


andsens:
Nice, an exemplary PR :-)
Thank you for the contribution John. And good on you for using the correct nomenclature wrt what `deb` and `deb-src` is called (i.e. "type").

Merging...

PR #205, opened 2015-03-27
Execute entries with a single string as shell commands

jwendell:
Documentation says about that, but it's not implemented like that.

If the command to be executed is an array with just one string, it's
likely the user passed the entire command inside the string.

Plus, they benefit from shell expansion of wildcards like '*'.

### DIFF
--- a/bootstrapvz/plugins/image_commands/README.md
+++ b/bootstrapvz/plugins/image_commands/README.md
@@ -10,6 +10,12 @@ Plugin is defined in the manifest file, plugin section with:
 
 The *commands* element is an array of commands. Each command is an array describing the executable and its arguments.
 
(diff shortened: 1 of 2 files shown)


andsens:
Yup , that makes a lot of sense. Nifty little enhancement :-)
We should either remove the README in the plugin folder and throw it in the wiki or do it the other way around with the plugins. But if you could copy paste the change into the wiki for now that'd be great.

Merging...

jwendell:
Is there a way to preview the doc page (gh-page branch) locally?

andsens:
Yup. Use gollum (https://github.com/gollum/gollum), it's what github uses to render the pages, though it won't be 1:1.

Issue #98, opened 2014-05-23
Load stand-alone plugins


myhro:
How can we load third-party plugins which are not part of the main bootstrap-vz repository? Anders and I have [discussed that in the beginning of the past month](https://github.com/andsens/bootstrap-vz/commit/33a430566b603f34e73e5be462822916e018283c#commitcomment-5901496), although I haven't yet looked at how they are currently being loaded. It would be nice if the plugins could be just proper python packages.

andsens:
Fixed with only 21 lines of code :-)

andsens:
Oh, and the docs are [here](http://bootstrap-vz.readthedocs.org/en/master/developers/plugins.html).

myhro:
Nicely done, Anders! The docs are also pretty comprehensive. :-)

PR #206, opened 2015-03-30
New plugin: ec2_launch

jwendell:
It adds the ability to automatically launch an EC2 instance
after AMI registration.

It has no mandatory configuration, only optional ones, like
instance type, security groups, etc. They should be documented in
later documentation patches.

### DIFF
--- a/bootstrapvz/plugins/ec2_launch/__init__.py
+++ b/bootstrapvz/plugins/ec2_launch/__init__.py
@@ -1,0 +1,11 @@
+def validate_manifest(data, validator, error):
+    import os.path
+    schema_path = os.path.normpath(os.path.join(os.path.dirname(__file__), 'manifest-schema.yml'))
+    validator(data, schema_path)
+
+
+def resolve_tasks(taskset, manifest):
+    import tasks
+    taskset.add(tasks.LaunchEC2Instance)
+    if 'print_public_ip' in manifest.plugins['ec2_launch']:
+        taskset.add(tasks.PrintPublicIPAddress)
--- a/bootstrapvz/plugins/ec2_launch/manifest-schema.yml
+++ b/bootstrapvz/plugins/ec2_launch/manifest-schema.yml
@@ -1,0 +1,18 @@
+---
+$schema: http://json-schema.org/draft-04/schema#
+title: EC2-launch plugin manifest
+type: object
+properties:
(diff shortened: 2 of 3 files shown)


andsens:
I feel I have to quote the contribution guidelines here:
> Please try to be very descriptive about your changes when you write a pull request, stating what it does, *why it is needed*, *which use cases this change covers* etc

If you are doing this with the motivation of testing the resulting AMI I'd recommend taking a look at the wip-integration-testing branch, which is intended to that and a lot more, so that one has a reasonable guarantee of the instance booting once the image is built.

jwendell:
In our environment we have people generating images [sometimes] various times a day, and, immediately after generating the AMI they go to launch it through EC2 console.

This plugin automates this step by launching an instance after the AMI is registered. [In our case, we should improve it to deregister the AMI and delete the snapshot].

So, this is our rationale to write this plugin. Sounds legit? Do you have other suggestion?

andsens:
Interesting. I still don't understand \*why* you do this. Is it to test the image or simply to have a one-off instance running?

jwendell:
We are in the middle of a heavy development of a debian-based product that will run on aws.

So, almost every developer/tester need to develop/test a feature on a fresh image, mostly generated from a variety of dev branches of our internal product.

I developed a build process (web/php frontend, python backend) that generates a manifest file on-the-fly and invokes bootstrap-vz. As a result of the process we have a running instance on AWS ready to be tested.

Just to enforce: This kind of test is not like 'is the instance booting normally?'. We are testing our product, not debian/bootstrap itself.

andsens:
Oooh. OK, that makes a lot more sense. Cool, at some point you might be able to use the unit testing framework I've built around bootstrap-vz in the branch I mentioned, but this seems like a good solution for you guys atm and I can see how this might be useful to others.

You should take a look at how [I am handling ec2 bootup, cleanup etc.](https://github.com/andsens/bootstrap-vz/blob/d35450ae3c0c8f20985cea64b21381270985ac05/tests/integration/images/ami.py), maybe there's something you can use to improve the plugin.

Once integration testing is done, I'll get working on 3rd party plugin integration (see #98). Then you wont have to wait for me to merge stuff to get things going or don't have to maintain a separate downstream branch that requires constant merging :-)

PR #207, opened 2015-04-02
ec2_launch: Allow to deregister the AMI after launching image

jwendell:
If all you want is to test an image or product and to achieve this
you need to generate several images a day, you will end up with
lots of AMI's and snapshots that have no use in the end of the day.

This commit adds the new boolean manifest option "deregister_ami"
that, if True, deletes the recently created AMI and snapshot. So, the
final result will be only the running instance, nothing else.

### DIFF
--- a/bootstrapvz/plugins/ec2_launch/__init__.py
+++ b/bootstrapvz/plugins/ec2_launch/__init__.py
@@ -9,3 +9,5 @@ def resolve_tasks(taskset, manifest):
     taskset.add(tasks.LaunchEC2Instance)
     if 'print_public_ip' in manifest.plugins['ec2_launch']:
         taskset.add(tasks.PrintPublicIPAddress)
+    if manifest.plugins['ec2_launch'].get('deregister_ami', False):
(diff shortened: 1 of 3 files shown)


andsens:
Woops. You got some PEP8 errors.
Heh, this plugin is more and more starting to look like my integration testing tool :-)

jwendell:
Sorry, I'm now running tox locally to prevent errors like this. Force-pushed the branch with the fix.

jwendell:
Hmm, please, do not merge it yet, it seems there's a race condition if I deregister the AMI immediately after launching the instance. I think I have to wait it to be in the running state, just to be sure.

I'll work on that and force-push the branch.

andsens:
Ah, ok. Hm, check out my integration testing branch, launch, deregistering etc. works perfectly there.

jwendell:
Done. I'm using the function waituntil() found on your integration-test branch. This allowed to simplify the code a little bit.

andsens:
Woops, forgot about this one. Merging...

PR #214, opened 2015-04-27
Add authentication support to the apt proxy plugin

jerrykan:
Add username and password settings to the APT Proxy plugin so that
users who are behind an authenticating proxy can still use bootstrap-vz
without having to jump through hoops. If either the username or password
are not set, then no authentication is used.

### DIFF
--- a/bootstrapvz/plugins/apt_proxy/tasks.py
+++ b/bootstrapvz/plugins/apt_proxy/tasks.py
@@ -34,11 +34,21 @@ class SetAptProxy(Task):
 	@classmethod
 	def run(cls, info):
 		proxy_path = os.path.join(info.root, 'etc/apt/apt.conf.d/02proxy')
+		proxy_username = info.manifest.plugins['apt_proxy'].get('username')
+		proxy_password = info.manifest.plugins['apt_proxy'].get('password')
 		proxy_address = info.manifest.plugins['apt_proxy']['address']
 		proxy_port = info.manifest.plugins['apt_proxy']['port']
+
+		if None not in (proxy_username, proxy_password):
+			proxy_auth = '{username}:{password}@'.format(
+				username=proxy_username, password=proxy_password)
+		else:
+			proxy_auth = ''
+
 		with open(proxy_path, 'w') as proxy_file:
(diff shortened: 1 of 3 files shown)


bootstrapvz/plugins/apt_proxy/tasks.py
```diff
@@ -34,11 +34,21 @@ class SetAptProxy(Task):
 	@classmethod
 	def run(cls, info):
 		proxy_path = os.path.join(info.root, 'etc/apt/apt.conf.d/02proxy')
+		proxy_username = info.manifest.plugins['apt_proxy'].get('username')
```
andsens: You will get a `KeyError` here if you do not supply a `None` as the default in the second parameter.

andsens:
Looks great. If you could fix the one issue I highlighted (and rebase it all), I'll merge. You are welcome to add yourself to the changelog btw.

jerrykan:
Thanks for looking at this. I don't believe it will raise a `KeyError`. From the python [docs](https://docs.python.org/2/library/stdtypes.html#dict.get):

> get(key[, default])
> Return the value for key if key is in the dictionary, else default. If default is not given, it defaults to None, so that this method never raises a KeyError.

andsens:
Oh wow, I had no clue. I always assumed `var['key']` was equivalent to `var.get('key')`.

Merging...

myhro:
> You will get a `KeyError` here if you do not supply a `None` as the default in the second parameter.

At first I thought you was talking about the `plugins['apt_proxy']` dict, but this will be checked only if the plugin is defined on the manifest, so there's no problem on both (`apt_proxy` and `username`) key access.

Nicely done, John. That was actually the commit number 1000 on the `development` branch. I've updated the changelog. :-)

andsens:
> That was actually the commit number 1000 on the development branch.

[Congratulations](http://i.imgur.com/gXGqIEw.gifv) @jerrykan :-)

jerrykan:
Thanks. Sorry it wasn't a more profound change ;)

I just discovered bootstrap-vz today and it is perfect for my needs. Thanks for putting the effort into making it.

PR #215, opened 2015-04-27
Enhancing admin_user plugin

DaniloMoura1:
Adding a few optional parameters to the <code>admin_user</code> plugin.

Also, noticed that it only disables root's ssh access if configured as <code>PermitRootLoin yes</code>


bootstrapvz/plugins/admin_user/tasks.py
```diff
@@ -69,7 +94,7 @@ def run(cls, info):
 			                'dpkg-query', '-W', 'openssh-server'])
 			from bootstrapvz.common.tools import sed_i
 			sshdconfig_path = os.path.join(info.root, 'etc/ssh/sshd_config')
-			sed_i(sshdconfig_path, 'PermitRootLogin yes', 'PermitRootLogin no')
+			sed_i(sshdconfig_path, 'PermitRootLogin (yes|without-password|forced-commands-only)', 'PermitRootLogin no')
```
andsens: Why would it be anything else than `PermitRootLogin yes`? It's a fresh Debian install we are modifying, there's no need to account for other configurations.

andsens:
Hm, I am not sure about this. Your code is good and the feature sensible, but I am a bit hesitant to merge because this is configuration management territory (and for that we have plugins supporting puppet, chef, salt and ansible).

It's a question of "where do we stop"? group memberships? location of the home folder? `.bashrc` contents? (and that's only for user configuration)

@myhro: What do you think?

myhro:
Anders,

The questions you raised are quite interesting. At first I thought that this patch is simple, so we could just merge it. But the nature of the modifications that it makes can not only be done by any orchestration tool - and we already have plenty support for them, but also done in a better and robust way.

andsens:
Exactly. OK, I think it's settled then. @DaniloMoura1 I'm closing this PR, sorry :-/
However, I'll pretty soon put some effort into solving #98, which would allow anyone to develop their own bootstrap-vz plugin, without having to merge it into this repo for it to work.

PR #216, opened 2015-04-28
root_password: Enable SSH root login

jwendell:
Jessie comes with root login disabled for passwords, however,
if the user is using this plugin, chances are high that they
need to login via SSH as root.


myhro:
Hi John. Thanks for sending the PR. This patch makes total sense to me, although I wasn't aware of the `jessie` defaults. What do you think about, @andsens ?

andsens:
Yup. Makes a lot of sense to me :-)
Merging...

Issue #218, opened 2015-05-08
There were 2 instead of 1 matches for the expression `PermitRootLogin .*' in the file `/target/2a279fd4/root/etc/ssh/sshd_config'


aderumier:
Hi,
since this commit
https://github.com/andsens/bootstrap-vz/commit/2d3a0a0ce35e17dc183cf353ace2aed84f20accc

setting ssh root password not working anymore,

There were 2 instead of 1 matches for the expression `PermitRootLogin .*' in the file `/target/2a279fd4/root/etc/ssh/sshd_config'

because
/etc/ssh/sshd_config

have 2 lines with PermitRootLogin (one of them is a comment)

# cat sshd_config |grep PermitRootLogin

PermitRootLogin yes
# the setting of "PermitRootLogin yes

The problem is that
sed_i(sshd_config_path, '^PermitRootLogin .*', 'PermitRootLogin yes')
has been replaced by
sed_i(sshdconfig_path, 'PermitRootLogin .*', 'PermitRootLogin yes')

(missing ^)

aderumier:
(this is with jesse)

andsens:
Thanks for the bug report. I'll fix it tomorrow if I have the time.
... or you can send a PR :-) Just remember that it still has to work for wheezy and squeeze.

myhro:
Closed by #219.

PR #219, opened 2015-05-08
fix ^PermitRootLogin regex

aderumier:
fix issue https://github.com/andsens/bootstrap-vz/issues/218

### DIFF
--- a/bootstrapvz/common/tasks/ssh.py
+++ b/bootstrapvz/common/tasks/ssh.py
@@ -59,7 +59,7 @@ class EnableRootLogin(Task):
 		sshdconfig_path = os.path.join(info.root, 'etc/ssh/sshd_config')
 		if os.path.exists(sshdconfig_path):
 			from bootstrapvz.common.tools import sed_i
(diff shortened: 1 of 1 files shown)


andsens:
Does this work on wheezy and squeeze? Is that line not commented out there?

myhro:
I guess this is an use case that fits for a test to avoid regressions. :-)

aderumier:
This is the same for squeeze/wheezy/jessie.

I never try bootstrap-vz with squeeze/wheezy, but the old regex before
https://github.com/andsens/bootstrap-vz/commit/2d3a0a0ce35e17dc183cf353ace2aed84f20accc
was using ^

myhro:
This patch works for `squeeze`/`wheezy`/`jessie` indeed. The line he was talking about in #218 is:

    (...)
    # the setting of "PermitRootLogin without-password".
    (...)

Without the `^`, this line got replaced too.

I'll be merging and adding you to the changelog, Alexandre. Thank you!

PR #223, opened 2015-06-04
Use SSD volumes instead of magnetic

jwendell:
They are faster and the default when using AWS EC2 console.

If you think this is too intrusive, we could add a config parameter to let the user choose the volume type he wants...


andsens:
Agreed. Merging :-)

PR #225, opened 2015-06-10
apt_proxy: Also consider http error code 406

jwendell:
It seems some Debian versions of apt-cacher-ng return error
code 406 instead of 404:

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=762969

This is my case in Jessie, for example.

This patch considers both codes as "valid".


andsens:
Ah yes, the good old 406 [Unacceptableeeee](https://www.youtube.com/watch?v=07So_lJQyqw).

PR #227, opened 2015-06-14
Ec2: Add the variable "install_init_scripts"

jwendell:
To control whether ec2 init scripts should be installed
on the image. Defaults to true, to keep current behaviour.


andsens:
Nice and small change, with documentation to boot :-)
Thanks John!

PR #228, opened 2015-06-17
EC2: Only install dhcpcd on releases before jessie

jwendell:
Otherwise, keep Debian's default isc-dhcp-client, it works fine
on EC2.


andsens:
Oh, how I love those PRs that are small and beyond that also remove more code than they add :-)

PR #229, opened 2015-06-18
Change the order we clean up the apt state for GCE images.

zmarano:
The result is that we leave a set of apt list files current as of the image build on the resulting image instead of no cache whatsoever.


andsens:
Gotcha, merging.

PR #230, opened 2015-06-19
Install Remote packages before Local ones.

wrigri:
Since Local packages are installed using dpkg, they do not automatically resolve
dependencies.  Installing remote packages first helps make sure that
dependencies that are included in the remote packages are taken care of by the
time the local packages are installed.


wrigri:
@zmarano

andsens:
Previously discussed and solved in andsens/build-debian-cloud#127. The package order in the manifest is significant, so you should be able to simply move the remote packages before the `.deb` and all will be fine. [I have now clarified this in the docs](http://bootstrap-vz.readthedocs.org/en/master/manifests/index.html#packages).

wrigri:
Fair enough, and I didn't see any issues with putting the packages in the correct order in the 'install:' section of the manifest.  What I did see, however, was that on occasion, other packages would get installed out-of-order.  For example, on one run, installing a .deb failed because the "sudo" package hadn't been installed yet.  It seems that it's possible for the packages in the manifest to be listed before some of the packages added by other tasks (although the exact order seems to be somewhat non-deterministic).  This PR may not be the best way to address that, but I'm not sure what the best way to address it is.

andsens:
Do you have the logs for any of those failed bootstraps by chance. It would be quite easy to glance the problem from the tasklist.

PR #232, opened 2015-06-22
Install GCE packages before Manifest packages.

wrigri:
After spending some additional time troubleshooting, this seems to be a better way to solve the issue that I was having with local packages defined in the manifest.


wrigri:
@zmarano

andsens:
Ah yes. Fair enough. Merging...

Issue #231, opened 2015-06-22
Azure provider fails installing the Azure agent for Linux


tomconte:
The process fails with the error below (incorrectly looking in 'system' section instead of 'provider'). Working on a fix.

```
'waagent'
Traceback (most recent call last):
  File "/home/azureuser/bootstrap-vz/bootstrapvz/base/main.py", line 109, in run
    tasklist.run(info=bootstrap_info, dry_run=dry_run)
  File "/home/azureuser/bootstrap-vz/bootstrapvz/base/tasklist.py", line 43, in run
    task.run(info)
  File "/home/azureuser/bootstrap-vz/bootstrapvz/providers/azure/tasks/packages.py", line 46, in run
    if info.manifest.system['waagent'].get('conf', False):
KeyError: 'waagent'
```

andsens:
Woops. That one slipped through the cracks when I refactored.
> Working on a fix.

Great! I'll hold off on fixing it myself then :-)

tomconte:
Fixed in b7243449f5264cf0da752e626a5a5392a7f39dc0

PR #233, opened 2015-06-23
Fix for issue #231 + added 'parted' package

tomconte:
- Use manifest.provider instead of manifest.system
- Add 'parted' package required for recent versions of Azure agent


andsens:
Looks great. Merging...

PR #234, opened 2015-06-24
Fix expand-root script to work for Debian 8 and simplify how partition types are formulated.

zmarano:
blkid output changed between wheezy and jessie breaking this script


andsens:
Yup. Much better way to do it. I use it that way for the grub config as well.

PR #235, opened 2015-06-26
Azure Jessie example manifest

tomconte:
Quick example manifest for Jessie on Azure

### DIFF
--- a/manifests/examples/azure/jessie.yml
+++ b/manifests/examples/azure/jessie.yml
@@ -4,7 +4,7 @@ provider:
   waagent:
     version: 2.0.13
 bootstrapper:
-  mirror: http://ftp.fr.debian.org/debian/
+  mirror: http://httpredir.debian.org/debian/
   workspace: /target
 image:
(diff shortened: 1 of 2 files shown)


myhro:
Hi Thomas,

Don't you think it would be nice to use a CDN or geographic mirror like http://httpredir.debian.org/ instead of the french one? The [wheezy manifest][1] is also using the same mirror.

Regards,
Tiago.

[1]: https://github.com/andsens/bootstrap-vz/blob/16b67f604bdc331b69be0b02fc19095773ab7a47/manifests/examples/azure/wheezy.yml

tomconte:
Sounds good to me! Fixed both manifests.

myhro:
Thomas,

I've forgot to mention that [it already defaults to it][1] if the `mirror` [key isn't present][2].

Thank you! Merging...

[1]: https://bootstrap-vz.readthedocs.org/en/master/manifests/index.html#packages
[2]: https://bootstrap-vz.readthedocs.org/en/master/manifests/index.html#bootstrapper

PR #236, opened 2015-06-29
Patch udev init script to fix ROOTDELAY boot issue

tomconte:
Hello guys, I would like to add a step to fix a boot delay issue with Azure. The `udev` init script is causing a delay when used with the recommended boot settings in Azure. (you have a link to the issue in the comments)
Feedback appreciated about the approach I took (applying a diff file instead of packing the complete file) -- overkill?
Thanks!

### DIFF
--- a/bootstrapvz/providers/azure/__init__.py
+++ b/bootstrapvz/providers/azure/__init__.py
@@ -2,6 +2,7 @@ from bootstrapvz.common import task_groups
 import tasks.packages
 import tasks.boot
 import tasks.image
+import tasks.udev
 from bootstrapvz.common.tasks import loopback
 from bootstrapvz.common.tasks import initd
 from bootstrapvz.common.tasks import ssh
@@ -15,7 +16,6 @@ def validate_manifest(data, validator, error):
 
 def resolve_tasks(taskset, manifest):
 	taskset.update(task_groups.get_standard_groups(manifest))
-
 	taskset.update([tasks.packages.DefaultPackages,
 	                loopback.AddRequiredCommands,
 	                loopback.Create,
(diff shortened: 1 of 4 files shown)


andsens:
> Feedback appreciated about the approach I took (applying a diff file instead of packing the complete file) -- overkill?

I like it. This more forward compatible than many other solutions. What do you think @myhro?

Before I merge, could you do a fixup on the formatting commit?

tomconte:
Fixup done.

andsens:
Merged. Thanks :-)

myhro:
Thank you, Thomas. :-)

Anders,

> I like it. This is more forward compatible than many other solutions. What do you think @myhro?

My only concern is if this patch applies on Debian versions other than `jessie`. Also, [shouldn't this][1] be `log_call()` or `log_check_call()`?

[1]: https://github.com/tomconte/bootstrap-vz/blob/774e78a9f25da147e49e48decb209b8d666513d7/bootstrapvz/providers/azure/tasks/udev.py#L19

andsens:
> My only concern is if this patch applies on Debian versions other than jessie.

Ah. Yeah, that might be a problem. Does wheezy currently work on azure @tomconte? And does the patch apply there?

> Also, shouldn't this be log_call() or log_check_call()?

Woops, totally overlooked that one. I'll fix it myself when I get home.

tomconte:
Yep wheezy should work and I also tested the diff on the wheezy version of that file. I will generate an image and double-check.

andsens:
I was wondering, why not just make sure that "$ROOTDELAY" is not set or 0? Wouldn't that be less invasive than patching a file?

tomconte:
Just checked that the patch works on the wheezy version of the file.
ROOTDELAY is typically passed as a parameter via the the Grub configuration. Typically for Azure we recommend rootdelay=300. This is intended as the maximum time to wait for the root device to become available, as sometimes this can take some time with a cloud drive.

Issue #237, opened 2015-06-29
EC2: Fixed volume size?


jwendell:
In the manifest file, I'm using a volume with 8G of size.
When I create an instance using the generated AMI, no matter the size of the storage I choose, the filesystem always will end up with 8G.

Is this the expected behavior?
In my opinion the volume should expand to fit the entire storage size...

Even the Debian official images "suffer" from this issue.

andsens:
Could you post the manifest and log as gists? Normally the root partition should be autoexpanded.

jwendell:
I think I found the issue:

cloud-init does that. Comparing to an ubuntu instance, I saw the cloud.cfg file is missing "growpart" in "cloud_init_modules" section.
In debian, the binary growpart is provided by cloud-utils package (which is not installed by default). I think it's a bit outdated, once cloud-init growpart module tries to access an invalid function of it: "growpart --update".

I'm trying to workaround this, but feel free to come with a solution :)

jwendell:
Ok, I copied ubuntu's growpart script to the debian box and it worked. So, to fix this, I guess we should:

1) Update cloud-utils package in Jessie
2) Install cloud-utils package in AddEC2InitScripts.run() (or, in step 1, make it a dependency of cloud-init)
3) Add the line "- growpart" right before "resizefs" in /etc/cloud/cloud.cfg (this might be done along with step 1)

flowhamster:
@jwendell ran into this while upgrading from wheezy to jessie AMI's using salt-cloud to provision. I'll spend some time later today to see if this will correct our issue.

myhro:
Closing, as this should have been closed when #269 was merged.

PR #239, opened 2015-06-30
EC2: Added growpart script

jwendell:
This is an attempt to fix issue #237

The script "growpart" was added as an asset, because
Debian's Jessie version (inside cloud-utils) package is
outdated and buggy.

Then "expand-root" init script was modified to call growpart
before calling resize2fs. In fact, calling resize2fs without
resizing the partition first is useless.

The task was named 'AddWorkaroundGrowpart' because it must
go away in the future in favor of using recent stuff in cloud-utils.

Currently even the official images suffer this issue, that makes
the system use only 8 GiB, even if the user chooses a bigger
storage size inside AWS.

### DIFF
--- a/bootstrapvz/common/assets/init.d/expand-root
+++ b/bootstrapvz/common/assets/init.d/expand-root
@@ -13,8 +13,19 @@
 prog=$(basename $0)
 logger="logger -t $prog"
 
-device_path="/dev/xvda"
+growpart=$(which growpart) || {
+	$logger "growpart was not found on PATH. Unable to expand size."
+	exit 0
+}
 
+root_index="0"
+
+$growpart /dev/xvda $root_index || {
+	$logger "growpart failed. Unable to expand size."
+	exit 0
+}
+
+device_path="/dev/xvda$root_index"
 filesystem=$(blkid -s TYPE -o value ${device_path})
 
 case $filesystem in
--- a/bootstrapvz/providers/ec2/__init__.py
+++ b/bootstrapvz/providers/ec2/__init__.py
@@ -67,6 +67,7 @@ def resolve_tasks(taskset, manifest):
 	                boot.DisableGetTTYs,
 	                tasks.network.EnableDHCPCDDNS,
 	                initd.AddExpandRoot,
+	                tasks.packages.AddWorkaroundGrowpart,
 	                initd.RemoveHWClock,
 	                initd.InstallInitScripts,
 
--- a/bootstrapvz/common/tasks/initd.py
+++ b/bootstrapvz/common/tasks/initd.py
@@ -57,9 +57,8 @@ class AdjustExpandRootScript(Task):
 
 	@classmethod
 	def run(cls, info):
-		import os.path
 		from ..tools import sed_i
 		script = os.path.join(info.root, 'etc/init.d/expand-root')
 		root_idx = info.volume.partition_map.root.get_index()
-		device_path = 'device_path="/dev/xvda{idx}"'.format(idx=root_idx)
-		sed_i(script, '^device_path="/dev/xvda"$', device_path)
+		root_index_line = 'root_index="{idx}"'.format(idx=root_idx)
+		sed_i(script, '^root_index="0"$', root_index_line)
--- a/bootstrapvz/providers/ec2/assets/bin/growpart
+++ b/bootstrapvz/providers/ec2/assets/bin/growpart
@@ -1,0 +1,587 @@
+#!/bin/sh
+#    Copyright (C) 2011 Canonical Ltd.
+#    Copyright (C) 2013 Hewlett-Packard Development Company, L.P.
+#
+#    Authors: Scott Moser <smoser@canonical.com>
+#             Juerg Haefliger <juerg.haefliger@hp.com>
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU General Public License as published by
+#    the Free Software Foundation, version 3 of the License.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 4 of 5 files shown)


jwendell:
Forgot to say: I'm *not* using cloud-init plugin.
If someone is using it, probably it's broken too, because there's no growpart script.

bootstrapvz/common/assets/init.d/expand-root
```diff
@@ -13,8 +13,19 @@
...
 
-device_path="/dev/xvda"
+growpart=$(which growpart) || {
+	$logger "growpart was not found on PATH. Unable to expand size."
+	exit 0
+}
 
+root_index="0"
+
+$growpart /dev/xvda $root_index || {
```
andsens: Wouldn't just calling `growpart` be fine here? We just established that it is on `$PATH`.

bootstrapvz/common/assets/init.d/expand-root
```diff
@@ -13,8 +13,19 @@
 prog=$(basename $0)
 logger="logger -t $prog"
 
-device_path="/dev/xvda"
+growpart=$(which growpart) || {
+	$logger "growpart was not found on PATH. Unable to expand size."
+	exit 0
```
andsens: Shouldn't we exit with something other than `0`?

bootstrapvz/common/assets/init.d/expand-root
```diff
@@ -13,8 +13,19 @@
 prog=$(basename $0)
 logger="logger -t $prog"
 
-device_path="/dev/xvda"
+growpart=$(which growpart) || {
```
andsens: We shouldn't use `which`. [Consider using `type` or `hash` instead](http://stackoverflow.com/a/677212/339505).

bootstrapvz/common/assets/init.d/expand-root
```diff
@@ -13,8 +13,19 @@
...
+}
 
+root_index="0"
+
+$growpart /dev/xvda $root_index || {
+	$logger "growpart failed. Unable to expand size."
+	exit 0
+}
+
+device_path="/dev/xvda$root_index"
```
andsens: I think it's better to allow changing the entire device path. This way we can easily switch to `/dev/by-uuid/` or something else if we want to.

bootstrapvz/providers/ec2/__init__.py
```diff
@@ -67,6 +67,7 @@ def resolve_tasks(taskset, manifest):
 	                boot.DisableGetTTYs,
 	                tasks.network.EnableDHCPCDDNS,
 	                initd.AddExpandRoot,
+	                tasks.packages.AddWorkaroundGrowpart,
```
andsens: Shouldn't the workaround task only be present when bootstrapping `>=jessie`?

andsens:
Sorry it took so long for me to answer this. July has been quite busy for me.
Could you check the comments I added?

bootstrapvz/common/assets/init.d/expand-root
jszwedko: Would exiting `1` work? Or did you have another code in mind.

bootstrapvz/common/assets/init.d/expand-root
jszwedko: Agreed :+1:

bootstrapvz/common/assets/init.d/expand-root
jszwedko: Also agreed :+1:

jszwedko:
Bump @jwendell @andsens I can pick up this work if needed -- we've been frustrated by this issue as well.

bootstrapvz/common/assets/init.d/expand-root
andsens: In my other project (homeshick) I have actually gathered [a list of exit codes](https://github.com/andsens/homeshick/blob/e8d021c10c7a54fba0a81a7995ca0148c07f3a9b/lib/exit_status.sh) from different sources. It seems `1` is the only one applicable.

bootstrapvz/common/assets/init.d/expand-root
andsens: In my other project (homeshick) I have actually gathered [a list of exit codes](https://github.com/andsens/homeshick/blob/e8d021c10c7a54fba0a81a7995ca0148c07f3a9b/lib/exit_status.sh) from different sources. It seems `1` is the only one applicable.

andsens:
Sure @jszwedko, that'd be fine by me.

bootstrapvz/common/assets/init.d/expand-root
jszwedko: :+1:

bootstrapvz/providers/ec2/__init__.py
jszwedko: Yeah, I think this is only applicable to `jessie`

jwendell:
Sure, go ahead :)

bootstrapvz/common/assets/init.d/expand-root
jszwedko: `root_index` is set above here -- did you envision `device_path` as a whole being configurable outside of this init script? Or just moving `/dev/xvda` to another variable here?

jszwedko: Ignore this, I see the `sed`ing happening below.

jszwedko: Is it possible to determine the disk and partition given the device path? I can `sed` just the device path into here, but `growpart` requires the disk and partition as arguments.

andsens: Just make it two variables :-)

andsens:
Closing, in favor of #269. Thank you for laying the foundation @jwendell! :+1:

PR #249, opened 2015-09-24
Fix a problem with Debian 8 on GCE with >2TB disks

wrigri:
This change makes two fixes. First, it fixes a race condition on Debian
8 GCE instances which caused instances with >2TB disks to fail. Second,
it adds the required "PREREQ" header to the script, which was previously
missing and caused errors at build time. These two issues are somewhat
unrelated to each other, but debugging the first led to finding the
second.


wrigri:
cc @zmarano

andsens:
Thank you :-)

Merged.

zmarano:
LGTM!

PR #250, opened 2015-09-25
GRUB_HIDDEN_TIMEOUT causes Debian 8 to not boot on GCE.

wrigri:


wrigri:
cc @zmarano

andsens:
Huh?! Any clue why, seems like an odd thing to affect the boot process.

wrigri:
I'm not sure, but I've been doing some additional research and I think that setting:
GRUB_HIDDEN_TIMEOUT=true
is the wrong thing to do.  Reading: http://www.gnu.org/software/grub/manual/grub.html#Configuration
it looks like what we actually may want is:
GRUB_HIDDEN_TIMEOUT=0
GRUB_HIDDEN_TIMEOUT_QUIET=true

According to the documentation, GRUB_HIDDEN_TIMEOUT should be set to the "Wait this many seconds for a key to be pressed before displaying the menu.", so I'm guessing that setting it to 'true' may be causing some undefined behavior.

I'll run some additional tests and let you know what I find out.

wrigri:
OK, I've tried
GRUB_HIDDEN_TIMEOUT=0
GRUB_HIDDEN_TIMEOUT_QUIET=true
on both Debian 7 and Debian 8 builds on GCE and they both worked as expected.

Obviously, I can make that change in just the GCE provider directory, but if you would prefer, I can send a PR to make that change in common instead.  Let me know how you would like to proceed.

andsens:
common. Definitely common, you've done your research and the current code is obviously wrong :-)

wrigri:
Withdrawing this PR in favor of #251

Issue #97, opened 2014-05-23
Remove disk image created by GCE after build


myhro:
The disk image used by the GCE provider isn't deleted after its tarball is generated. We should remove that file by the end of the build.

PR #254, opened 2015-10-16
Make generation of an image tarball a common task

myhro:
@andsens there are a couple things here which I would like to ask for your opinion. :-)

[Changes 'MoveImage' from `loopback` to `image`][1]

Moved `MoveImage` from the `loopback` common module a new one named `image`. Does that sounds reasonable to you? To me, it makes more sense if it grouped with tasks regarding images, not exactly loopback volumes.

[Make generation of an image tarball a common task][2]

I've created a common task to generate a compressed tarball of the image file, heavily based on a GCE task which did exactly this. Two problems appeared with this approach:

* What is the proper way to add a new attribute to the `info` object? I've tried to [create a new dict][3], but if anyone wants to add some other parameters to it on another task, they might be overwritten. This is defined using a [simple variable attribution][4] at some places, but this looks like to work only with attributes which their name starts with underscore. Possibly, there are attributes like this being generated for every provider, e.g. `gce` has a `info._gce`?
* The [image tarball name generation for GCE][5] is a little bit complex. At first, I thought that was just a matter of defining the name on the manifest itself, but this isn't possible. GCE is using [a specific task][6] to gather information about the Debian version that is being bootstrapped, then using this data to generate the image tarball name. I've though about creating an intermediary task which defines the image name (is it possible to have a generic one, then overriding it when needed for a specific to the provider? Or do we have to manually discard the generic and add the specific to the tasklist?), and then checking if the name was already defined (and stored on the `info` object). Do you see a simpler way to do this?
* This is probably more suited to be answered by @wrigri or @zmarano: does the RAW image file, stored on the compressed image tarball **should** be named `disk.raw` on GCE?

[1]: https://github.com/andsens/bootstrap-vz/commit/69c5f2fbec854fb9a4e038bf31351a23a16c0ab9
[2]: https://github.com/andsens/bootstrap-vz/commit/0b5a0737d4d0068acc47ffb4fedfe1aaf58db152
[3]: https://github.com/andsens/bootstrap-vz/commit/0b5a0737d4d0068acc47ffb4fedfe1aaf58db152#diff-c6be75ec30df38d0db25e863bcbba1daR44
[4]: https://github.com/andsens/bootstrap-vz/blob/0068cdc1a8a15061856a962045984126476b9583/bootstrapvz/providers/gce/tasks/image.py#L28
[5]: https://github.com/andsens/bootstrap-vz/blob/0068cdc1a8a15061856a962045984126476b9583/bootstrapvz/providers/gce/tasks/image.py#L20
[6]: https://github.com/andsens/bootstrap-vz/blob/0068cdc1a8a15061856a962045984126476b9583/bootstrapvz/providers/gce/tasks/configuration.py#L6

### DIFF
--- a/bootstrapvz/base/bootstrapinfo.py
+++ b/bootstrapvz/base/bootstrapinfo.py
@@ -55,6 +55,13 @@ class BootstrapInformation(object):
 		# that hint at how a command may be made available.
 		self.host_dependencies = {}
 
+		# Dictionary to store information about the image, like its resulting name
+		image_name = self.__parse_image_name(self.manifest, self.manifest_vars)
+		self.image = {
+		    'filename': image_name + '.' + self.volume.extension,
+		    'name': image_name,
+		}
+
 		# Lists of startup scripts that should be installed and disabled
 		self.initd = {'install': {}, 'disable': []}
 
@@ -130,6 +137,12 @@ class BootstrapInformation(object):
 		state = filter_state(self.__dict__)
 		state['__class__'] = self.__module__ + '.' + self.__class__.__name__
 		return state
+
+	def __parse_image_name(self, manifest, manifest_vars):
+		image_name = manifest.image['name'].format(**manifest_vars)
+		# Ensure that we do not use disallowed characters in image name
+		image_name = image_name.lower().replace('.', '-').replace(' ', '-')
+		return image_name
 
 	def __setstate__(self, state):
 		for key in state:
--- a/bootstrapvz/common/tasks/image.py
+++ b/bootstrapvz/common/tasks/image.py
@@ -10,10 +10,7 @@ class MoveImage(Task):
 
 	@classmethod
 	def run(cls, info):
-		image_name = info.manifest.image['name'].format(**info.manifest_vars)
-		filename = image_name + '.' + info.volume.extension
-
-		destination = os.path.join(info.manifest.bootstrapper['workspace'], filename)
+		destination = os.path.join(info.manifest.bootstrapper['workspace'], info.image['filename'])
 		import shutil
 		shutil.move(info.volume.image_path, destination)
 		info.volume.image_path = destination
@@ -29,23 +26,12 @@ class CreateImageTarball(Task):
 
 	@classmethod
 	def run(cls, info):
-		image_name = info.manifest.image['name'].format(**info.manifest_vars)
-		filename = image_name + '.' + info.volume.extension
+		if not info.image.get('tarball_name', False):
+			tarball_name = info.image['name'] + '.tar.gz'
+			info.image['tarball_name'] = tarball_name
 
-		# Ensure that we do not use disallowed characters in image name
-		image_name = image_name.lower()
-		image_name = image_name.replace('.', '-')
-		image_name = image_name.replace(' ', '-')
-
-		tarball_name = image_name + '.tar.gz'
-		tarball_path = os.path.join(info.manifest.bootstrapper['workspace'], tarball_name)
-
-		# Store image information so it can be referenced later
-		info.image = {
-		    'name': image_name,
-		    'tarball_name': tarball_name,
-		    'tarball_path': tarball_path,
-		}
+		tarball_path = os.path.join(info.manifest.bootstrapper['workspace'], info.image['tarball_name'])
+		info.image['tarball_path'] = tarball_path
 
 		log_check_call(['tar', '--sparse', '-C', info.manifest.bootstrapper['workspace'],
-		                '-caf', tarball_path, filename])
+		                '-caf', tarball_path, info.image['filename']])
--- a/bootstrapvz/common/tasks/image.py
+++ b/bootstrapvz/common/tasks/image.py
@@ -1,5 +1,7 @@
 from bootstrapvz.base import Task
 from bootstrapvz.common import phases
+from bootstrapvz.common.tools import log_check_call
+import os
 
 
 class MoveImage(Task):
@@ -11,7 +13,6 @@ class MoveImage(Task):
 		image_name = info.manifest.image['name'].format(**info.manifest_vars)
 		filename = image_name + '.' + info.volume.extension
 
-		import os.path
 		destination = os.path.join(info.manifest.bootstrapper['workspace'], filename)
 		import shutil
 		shutil.move(info.volume.image_path, destination)
@@ -19,3 +20,32 @@ class MoveImage(Task):
 		import logging
 		log = logging.getLogger(__name__)
 		log.info('The volume image has been moved to ' + destination)
+
+
+class CreateImageTarball(Task):
+	description = 'Creating tarball with image'
+	phase = phases.image_registration
+	predecessors = [MoveImage]
+
+	@classmethod
+	def run(cls, info):
+		image_name = info.manifest.image['name'].format(**info.manifest_vars)
+		filename = image_name + '.' + info.volume.extension
+
+		# Ensure that we do not use disallowed characters in image name
+		image_name = image_name.lower()
+		image_name = image_name.replace('.', '-')
+		image_name = image_name.replace(' ', '-')
+
+		tarball_name = image_name + '.tar.gz'
+		tarball_path = os.path.join(info.manifest.bootstrapper['workspace'], tarball_name)
+
+		# Store image information so it can be referenced later
+		info.image = {
+		    'name': image_name,
+		    'tarball_name': tarball_name,
+		    'tarball_path': tarball_path,
+		}
+
+		log_check_call(['tar', '--sparse', '-C', info.manifest.bootstrapper['workspace'],
+		                '-caf', tarball_path, filename])
--- a/bootstrapvz/providers/gce/tasks/image.py
+++ b/bootstrapvz/providers/gce/tasks/image.py
@@ -2,47 +2,17 @@ from bootstrapvz.base import Task
 from bootstrapvz.common import phases
 from bootstrapvz.common.tasks import image
 from bootstrapvz.common.tools import log_check_call
-import os.path
-
-
-class CreateTarball(Task):
-	description = 'Creating tarball with image'
-	phase = phases.image_registration
-	predecessors = [image.MoveImage]
-
-	@classmethod
-	def run(cls, info):
-		import datetime
-		image_name = info.manifest.image['name'].format(**info.manifest_vars)
-		filename = image_name + '.' + info.volume.extension
-		today = datetime.datetime.today()
-		name_suffix = today.strftime('%Y%m%d')
-		image_name_format = '{lsb_distribution}-{lsb_release}-{release}-v{name_suffix}'
-		image_name = image_name_format.format(lsb_distribution=info._gce['lsb_distribution'],
-		                                      lsb_release=info._gce['lsb_release'],
-		                                      release=info.manifest.system['release'],
-		                                      name_suffix=name_suffix)
-		# ensure that we do not use disallowed characters in image name
-		image_name = image_name.lower()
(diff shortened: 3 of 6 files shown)


zmarano:
For GCE the raw file must be named disk.raw so ideally you can have an override for providers if you're going to make this common.

myhro:
Thanks for the feedback, Zach! I'll take care to not break the GCE with this change.

andsens:
Hi @myhro.

> Moved MoveImage from the loopback common module a new one named image.

Agreed. Makes sense.

> What is the proper way to add a new attribute to the info object? I've tried to create a new dict, but if anyone wants to add some other parameters to it on another task, they might be overwritten.

So far [I have added them when creating the BootstrapInfo object](https://github.com/andsens/bootstrap-vz/blob/0068cdc1a8a15061856a962045984126476b9583/bootstrapvz/base/bootstrapinfo.py#L59). It's not an optimal solution, since some things may be very provider/plugin specific.

Adding an image attribute seems like something that is generally useful, so I'd add it there.  
Existing attributes with info about the image would have to be changed to use that attribute, in order to keep things consistent.

> I've though about creating an intermediary task which defines the image name

Why is this necessary at all? Don't we have `image.name` in the manifest for that? If it's a matter of too few variables to choose from, we can always add more :-)

> Or do we have to manually discard the generic and add the specific to the tasklist?

Definitely this. You want to avoid confusion when reading through the task list. One task should never undo another tasks work, that can get confusing very quickly (maybe this should be added to the contrib guidelines?).

p.s.: The code itself is top notch of course, so I assume you'll just merge when you think it's ready.

myhro:
> So far [I have added them when creating the BootstrapInfo object](https://github.com/andsens/bootstrap-vz/blob/0068cdc1a8a15061856a962045984126476b9583/bootstrapvz/base/bootstrapinfo.py#L59). It's not an optimal solution, since some things may be very provider/plugin specific.

Even checking the `bootstrapinfo.py` file before asking this, I hadn't seen those lines which perfectly answers my question. :-)

> Adding an image attribute seems like something that is generally useful, so I'd add it there.

Agreed. I'll add it there.

> Existing attributes with info about the image would have to be changed to use that attribute, in order to keep things consistent.

No problem. I'll look for them.

> Why is this necessary at all? Don't we have `image.name` in the manifest for that? If it's a matter of too few variables to choose from, we can always add more :-)

Mostly because of use cases like the one that I mentioned about GCE, when its RAW volume name differs from the image tarball name. I'm still not seeing a way this can be implemented in a clear and simple way, like adding a new variable in the manifest.

At one side, the [current `name` for GCE images (`disk`)][1] means nothing to anyone who reads the manifest. [I had some troubles myself][2] when it wasn't deleting the volume after generating the tarball, not knowing where their names diverged.

On the other side, having a `name` and `tarball_name` for the `image` manifest's section can be a little bit confusing. Of course, this is always something we can explain on the documentation, but I worry about losing a bit of usability.

> Definitely this. You want to avoid confusion when reading through the task list. One task should never undo another tasks work, that can get confusing very quickly (maybe this should be added to the contrib guidelines?).

Oh, sorry. You already told me this some time ago and I forgot about it. I totally agree! We should definitely add a note about this to the contribution guidelines.

> p.s.: The code itself is top notch of course, so I assume you'll just merge when you think it's ready.

Thank you, Anders. Actually, this PR is mostly a Request-For-Comments, as it is breaking the GCE implementation on its current form. I'll fix those problems and update it.

[1]: https://github.com/andsens/bootstrap-vz/blob/0068cdc1a8a15061856a962045984126476b9583/manifests/official/gce/jessie.yml#L7
[2]: https://github.com/andsens/bootstrap-vz/issues/97

myhro:
Anders, I've spent some time discussing with @fernandomrm (which I would like to thank a lot for the pair programming session that lasted the entire afternoon!) about the possibility of creating the image tarball name from a new variable added to the manifest. We came to the conclusion that this will not be simple as defining a specific task to do this, mostly because it could be using variables that are defined elsewhere and not on the manifest itself (so `'...'.format(**info.manifest_vars)` wouldn't be enough for these cases).

Could you please take a look at the PR's current state? GCE is not being broke anymore and we used the `info.image` attribute to avoid image name generation code being duplicated at a few places.

bootstrapvz/base/bootstrapinfo.py
```diff
@@ -131,6 +138,12 @@ def filter_state(state):
 		state['__class__'] = self.__module__ + '.' + self.__class__.__name__
 		return state
 
+	def __parse_image_name(self, manifest, manifest_vars):
+		image_name = manifest.image['name'].format(**manifest_vars)
+		# Ensure that we do not use disallowed characters in image name
+		image_name = image_name.lower().replace('.', '-').replace(' ', '-')
```
andsens: I don't think this should happen here. The allowed character set is very provider specific - I suggest enforcing allowable characters through the manifest schema instead, this way you wont surprise users either when they are looking for the completed image.

andsens:
> We came to the conclusion that this will not be simple as defining a specific task to do this, mostly because it could be using variables that are defined elsewhere and not on the manifest itself (so '...'.format(**info.manifest_vars) wouldn't be enough for these cases).

You can always add more variables specifically for the image name that are not a part of the manifest vars. Just merge the manifest vars and the new vars into a new dict before passing them to `format()`.

bootstrapvz/base/bootstrapinfo.py
```diff
@@ -55,6 +55,13 @@ def __init__(self, manifest=None, debug=False):
 		# that hint at how a command may be made available.
 		self.host_dependencies = {}
 
+		# Dictionary to store information about the image, like its resulting name
+		image_name = self.__parse_image_name(self.manifest, self.manifest_vars)
+		self.image = {
+		    'filename': image_name + '.' + self.volume.extension,
```
andsens: EBS backed volumes don't have an extension and the vagrant plugin uses `.box` as an extension, which is not a volume in itself. It it wise to lock in those things here? Wouldn't it be better to just define the dict and let other tasks handle the dict population?

bootstrapvz/common/tasks/image.py
```diff
@@ -0,0 +1,37 @@
...
+		destination = os.path.join(info.manifest.bootstrapper['workspace'], info.image['filename'])
+		import shutil
+		shutil.move(info.volume.image_path, destination)
+		info.volume.image_path = destination
+		import logging
+		log = logging.getLogger(__name__)
+		log.info('The volume image has been moved to ' + destination)
+
+
+class CreateImageTarball(Task):
```
andsens: Besides GCE, which provider uses this task?

bootstrapvz/providers/gce/tasks/image.py
```diff
@@ -56,6 +26,26 @@ def run(cls, info):
 		if 'description' in info.manifest.image:
 			image_description = info.manifest.image['description']
 		log_check_call(['gcutil', '--project=' + info.manifest.image['gce_project'],
-		                'addimage', info._gce['image_name'],
-		                info.manifest.image['gcs_destination'] + info._gce['tarball_name'],
+		                'addimage', info.image['name'],
+		                info.manifest.image['gcs_destination'] + info.image['tarball_name'],
```
andsens: If the gce provider uploads the image itself, why does the user need to be able to specify the tarball name? If this were to run fully automatic, we could essentially delete the tarball after a successful run, right?

bootstrapvz/providers/gce/tasks/image.py
```diff
@@ -56,6 +26,26 @@ def run(cls, info):
...
+	phase = phases.image_registration
+	successors = [image.CreateImageTarball]
+
+	@classmethod
+	def run(cls, info):
+		import datetime
+		today = datetime.datetime.today()
+		name_suffix = today.strftime('%Y%m%d')
+		image_name_format = '{lsb_distribution}-{lsb_release}-{release}-v{name_suffix}'
+		image_name = image_name_format.format(lsb_distribution=info._gce['lsb_distribution'],
```
andsens: It would be cool if we could pre-computer the `lsb*` vars and add them to the manifest vars. On the other hand: https://en.wikipedia.org/wiki/Linux_Standard_Base#Limitations_on_Debian

bootstrapvz/base/bootstrapinfo.py
myhro: Indeed. I guess it's better if we postpone the image name generation. But there's another problem here:

* In GCE, there are two distinct names: the image volume name, `disk.raw`, and the image tarball name, e.g.: `debian-8-2-jessie-v20151017.tar.gz`.
* The Vagrant plugin doesn't cares about the image volume name, as it uses [a symbolic link][1] to "rename" it to the expected `box-disk1.*`. Then, [everything is packed][2] in a `.box` file (which is also a gzipped tarball). Maybe we should use a similar approach for GCE?

[1]: https://github.com/andsens/bootstrap-vz/blob/0068cdc1a8a15061856a962045984126476b9583/bootstrapvz/plugins/vagrant/tasks.py#L134
[2]: https://github.com/andsens/bootstrap-vz/blob/0068cdc1a8a15061856a962045984126476b9583/bootstrapvz/plugins/vagrant/tasks.py#L142

bootstrapvz/base/bootstrapinfo.py
myhro: This is another reason why your suggestion of dealing with this at the manifest level would be better. :-)

bootstrapvz/common/tasks/image.py
myhro: Oh, I forgot to mention: this is also a requirement for Oracle Cloud. That's the reason we need to make it a common task.

bootstrapvz/providers/gce/tasks/image.py
myhro: This step is optional: the tarball is only uploaded [if the user wants to][1]. Also, its name isn't really defined by the user, but by [this definition][2], which was made (or at least commited) by @rybaktomasz.

[1]: https://github.com/andsens/bootstrap-vz/blob/0068cdc1a8a15061856a962045984126476b9583/bootstrapvz/providers/gce/__init__.py#L62
[2]: https://github.com/andsens/bootstrap-vz/blob/0068cdc1a8a15061856a962045984126476b9583/bootstrapvz/providers/gce/tasks/image.py#L8

bootstrapvz/providers/gce/tasks/image.py
myhro: I've read about [Debian dropping the Linux Standard Base][1] yesterday, but I'm not sure what we could do here. The **only** LSB information used here by GCE that isn't available at the manifest is the `major.minor` release number, e.g.: `debian-8-2-jessie-v20151017.tar.gz` (`8-2` is the replaced `8.2`). We can get the major version from [our `releases` module][2], but what about the minor?

[1]: https://lwn.net/Articles/658809/
[2]: https://github.com/andsens/bootstrap-vz/blob/0068cdc1a8a15061856a962045984126476b9583/bootstrapvz/common/releases.py#L34

myhro:
I've merged the change regarding `MoveImage` and renamed the PR, as it is now focusing at the image tarball generation.

andsens:
Just to make sure. Do you intend to merge this as is? Because right now `__parse_image_name()` is still in this PR.

myhro:
No no. I won't merge it until we agree with its current state. Do you prefer to close it while I work on my fork?

andsens:
Ah. No don't worry about it. I just misunderstood your previous comment :-)

myhro:
Anders,

I'm closing this PR as it has many conflicts right now. I'll use our discussion here as a basis for a new approach to make `CreateImageTarball` a common task.

andsens:
Alright, cool :-)

PR #255, opened 2015-10-23
Remove support for the GCE Debian mirror.

illfelder:


myhro:
Hi Max,

The GCE mirror was (or is going to be) discontinued?

illfelder:
We are starting the deprecation process for the GCE Debian mirror - removing references to it in future images is the first step.

myhro:
Thanks for the clarification. Merging... :-)

illfelder:
Awesome! Thanks!

Issue #144, opened 2014-09-02
install_standard option does not work on squeeze/oldstable


myhro:
I was trying to reproduce [this grub error](https://github.com/andsens/bootstrap-vz/issues/122#issue-37736340), when an unrelated issue was raised:

    Adding standard packages from tasksel
    Executing: chroot /target/f76d3ce9/root tasksel --task-packages standard
    ~pstandard
    ~prequired
    ~pimportant
    Upgrading packages and fixing broken dependencies
    Executing: chroot /target/f76d3ce9/root apt-get install --fix-broken --no-install-recommends --assume-yes
    Reading package lists...
    Building dependency tree...
    0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
    Executing: chroot /target/f76d3ce9/root apt-get upgrade --no-install-recommends --assume-yes
    Reading package lists...
    Building dependency tree...
    0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
    Installing packages
    Executing: chroot /target/f76d3ce9/root apt-get install --no-install-recommends --assume-yes openssh-server linux-image-amd64 grub-pc sudo nfs-client ~pstandard ~prequired ~pimportant
    Reading package lists...
    Building dependency tree...
    E: Unable to locate package ~pstandard
    E: Unable to locate package ~prequired
    E: Unable to locate package ~pimportant
    apt exited with status code 100. This can sometimes occur when package retrieval times out or a package extraction failed. apt might succeed if you try bootstrapping again.
    Command 'chroot /target/f76d3ce9/root apt-get install --no-install-recommends --assume-yes openssh-server linux-image-amd64 grub-pc sudo nfs-client ~pstandard ~prequired ~pimportant' returned non-zero exit status 100
    Traceback (most recent call last):
      File "/vagrant/bootstrapvz/base/main.py", line 109, in run
        tasklist.run(info=bootstrap_info, dry_run=dry_run)
      File "/vagrant/bootstrapvz/base/tasklist.py", line 38, in run
        task.run(info)
      File "/vagrant/bootstrapvz/common/tasks/packages.py", line 38, in run
        actions[package.__class__](info, batch)
      File "/vagrant/bootstrapvz/common/tasks/packages.py", line 73, in install_remote
        raise e
[...]

Issue #122, opened 2014-07-13
For how long we should continue supporting squeeze/oldstable?


myhro:
I was testing a fix for [this problem with backports syntax](https://github.com/andsens/bootstrap-vz/issues/96#issuecomment-48819079) in `squeeze`/`oldstable` that I reported on #96 and got a `CalledProcessError: Command 'chroot /target/5e231a2a/root grub-install /dev/dm-0' returned non-zero exit status 1`. Before trying to investigate this problem with `grub-install`, I started to ask myself for how long we will be worrying in providing fixes for this release.

The official support for `squeeze`/`oldstable` ended on last May 31st. Probably isn't news for anyone, but its [Long Term Support (LTS)](https://wiki.debian.org/LTS) was announced to last until Feb/2016, which is a little bit more than 18 months ahead. I see that makes sense for us to support it, as `squeeze-lts` targets [exactly the same two architectures that we do](https://wiki.debian.org/LTS/Using#Add_squeeze-lts_to_your_sources.list), `amd64` and `i386`, but is there a real need for that? To me, looks like that the public that this LTS support targets is not the "cloud public" that we do.

I would like to hear not only what Anders has to say about this, but also James (@JamesBromberger), Jimmy (@jkaplowitz), Olivier (@osallou) and Tomasz (@rybaktomasz) - and, of course, everyone else who would be interested in the matter.

Ps.: maybe I should also ask this on `debian-cloud` mailing list?

myhro:
I'm closing this issue, as the decision has already been made.

PR #265, opened 2016-01-07
enable support for creating kvm images for jessie on arm64

clarktlaugh:
These minor changes allow bootstrap-vz to build a kvm image for jessie on arm64, by updating the schema to accept arm64 as an architecture, no bootloader, and providing an example manifest.

### DIFF
--- a/bootstrapvz/base/manifest-schema.yml
+++ b/bootstrapvz/base/manifest-schema.yml
@@ -42,7 +42,7 @@ properties:
     type: object
     properties:
       architecture:
-        enum: [i386, amd64]
+        enum: [i386, amd64, arm64]
       userspace_architecture:
         enum: [i386]
       bootloader:
--- a/bootstrapvz/providers/kvm/manifest-schema.yml
+++ b/bootstrapvz/providers/kvm/manifest-schema.yml
@@ -26,6 +26,7 @@ properties:
         enum:
         - grub
         - extlinux
+        - none
   volume:
     type: object
     properties:
--- a/bootstrapvz/providers/kvm/tasks/packages-kernels.yml
+++ b/bootstrapvz/providers/kvm/tasks/packages-kernels.yml
@@ -9,6 +9,7 @@ wheezy:
 jessie:
   amd64: linux-image-amd64
   i386: linux-image-686-pae
+  arm64: linux-image-arm64
 sid:
   amd64: linux-image-amd64
   i386: linux-image-686-pae
--- a/manifests/examples/kvm/jessie-arm64-virtio.yml
+++ b/manifests/examples/kvm/jessie-arm64-virtio.yml
@@ -1,0 +1,32 @@
+---
+name: debian-{system.release}-{system.architecture}-{%y}{%m}{%d}
+provider:
+  name: kvm
+  virtio_modules:
+  - virtio_pci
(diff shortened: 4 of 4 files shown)


myhro:
Hi Clark,

I don't have a way to test an `arm64` image (maybe on QEMU?) right now. It is nice to know that a new architecture can be bootstrapped with so few changes, but I'm not really sure if we have enough resources (e.g.: manpower, time and hardware) to maintain it.

What are your thoughts on this, Anders?

clarktlaugh:
Yes - QEMU should work fine for testing an image.

I work with Linaro, which is building an arm64 reference platform based on Debian  (https://github.com/96boards/documentation/wiki/ReferenceSoftwareEE) with a big focus on Openstack and docker.  Having the standard tooling support creation of proper images will be a priority, so I am sure help with testing/maintenance would not be an issue.  But, hopefully, the impact to bootstrap-vz would be minimal?

andsens:
@myhro 

> but I'm not really sure if we have enough resources (e.g.: manpower, time and hardware) to maintain it.

It's a valid concern. It seems, however, that most of the legwork is done by Debian itself. bootstrap-vz is quite architecture agnostic really. So in the end, I don't think merging this PR would increase the workload.

@clarktlaugh 

> Having the standard tooling support creation of proper images will be a priority, so I am sure help with testing/maintenance would not be an issue.

That would be awesome. bootstrap-vz has [an entire testing harness](http://bootstrap-vz.readthedocs.org/en/master/testing/index.html) to help you do that. A new integration test provider [is quite easy to create](http://bootstrap-vz.readthedocs.org/en/master/testing/integration_test_providers/index.html#implementation).

I'm fine with merging this, @myhro (merge if you agree :-) )?

myhro:
Anders,

> So in the end, I don't think merging this PR would increase the workload.

My only concern here are those corner cases where fixing a problem in an architecture would break another one - specially if this isn't easy to test on many releases/providers. Maybe we should state that our goal is to provide support for the current `stable` release on architectures `x`, `y` and `z` (and some of them only on a subset of providers, e.g.: `arm64` only on `kvm`) and everything else is maintained as a "best effort"?

P.s.: this is something that has been on my head since #122 / #144. :-(

> I'm fine with merging this, @myhro? (merge if you agree :-) )

I would like to see those commits squashed and `jessie-virtio.yml` to be renamed mentioning its architecture (like `jessie-arm64-virtio.yml`) to avoid confusion (even with the name `arm64` being pretty close to `amd64`). Could you please do this, @clarktlaugh?

andsens:
@myhro, you're head is definitely more in the game than mine is right now :-)
Agree with all your points.

> Maybe we should state that our goal is to provide support for the current stable release on architectures

Yup. Hmm, sounds like an addition to [Supported builds](http://bootstrap-vz.readthedocs.org/en/master/supported_builds.html), we should link to that from somewhere in the top of the readme.

myhro:
Clark,

Sorry for the delay, but GitHub doesn't notify about new commits on a Pull Request - only comments. I'm merging this one. Thanks a lot!

-

Anders,

> Yup. Hmm, sounds like an addition to Supported builds, we should link to that from somewhere in the top of the readme.

I've opened #266 so we can track this separately.

PR #256, opened 2015-11-05
EC2 update: Systemd, module blacklisting, and nofail in fstab for cloud-init

JamesBromberger:

PR #276, opened 2016-01-29
Add support for Stretch, Virtualbox

apolloclark:
This attempts to disable "Persistent Network Interface Naming", but still requires running "sudo update-grub" to fully activate.

### DIFF
--- a/bootstrapvz/base/manifest-schema.yml
+++ b/bootstrapvz/base/manifest-schema.yml
@@ -55,7 +55,6 @@ properties:
         type: object
         properties:
           enable_timeout: {type: integer}
-          disable_pnin: {type: boolean}
         additionalProperties: false
       charmap: {type: string}
       hostname:
--- a/bootstrapvz/base/manifest-schema.yml
+++ b/bootstrapvz/base/manifest-schema.yml
@@ -45,14 +45,18 @@ properties:
         enum: [i386, amd64, arm64]
       userspace_architecture:
         enum: [i386]
-      grub_enable_timeout: {type: boolean}
-      grub_disable_pnin: {type: boolean}
       bootloader:
         enum:
         - pvgrub
         - grub
         - extlinux
         - none
+      grub:
+        type: object
+        properties:
+          enable_timeout: {type: integer}
+          disable_pnin: {type: boolean}
+        additionalProperties: false
       charmap: {type: string}
       hostname:
         type: string
--- a/bootstrapvz/base/manifest-schema.yml
+++ b/bootstrapvz/base/manifest-schema.yml
@@ -45,6 +45,8 @@ properties:
         enum: [i386, amd64, arm64]
       userspace_architecture:
         enum: [i386]
+      grub_enable_timeout: {type: boolean}
+      grub_disable_pnin: {type: boolean}
       bootloader:
         enum:
         - pvgrub
--- a/bootstrapvz/common/tasks/grub.py
+++ b/bootstrapvz/common/tasks/grub.py
@@ -25,26 +25,59 @@ class ConfigureGrub(Task):
 	def run(cls, info):
 		from bootstrapvz.common.tools import sed_i
 		grub_def = os.path.join(info.root, 'etc/default/grub')
+
+		print dir(info.manifest.system)
+		print info.manifest.system
+		print dir(info.manifest.system['grub'])
+		print info.manifest.system['grub']
+
+		# disable the graphical grub menu
 		sed_i(grub_def, '^#GRUB_TERMINAL=console', 'GRUB_TERMINAL=console')
-		
-		# optionally, enable grub menu
-		if info.manifest.system.get('grub_enable_timeout') is True:
-			# print "grub: enabling menu"
-			sed_i(grub_def, '^GRUB_TIMEOUT=[0-9]+', 'GRUB_TIMEOUT=5')
+
+		# optional grub settings
+		if 'grub' in info.manifest.system:
+			grub_settings = info.manifest.system['grub']
+
+			# enable grub menu
+			if grub_settings.get('enable_timeout') > 0:
+				print "grub: enabling timeout"
+				sed_i(
+					grub_def,
+					'^GRUB_TIMEOUT=[0-9]+',
+					'GRUB_TIMEOUT=' + str(grub_settings.get('enable_timeout'))
+				)
+			else:
+				sed_i(
+					grub_def,
+					'^GRUB_TIMEOUT=[0-9]+',
+					'GRUB_TIMEOUT=0\n'
+					'GRUB_HIDDEN_TIMEOUT=0\n'
+					'GRUB_HIDDEN_TIMEOUT_QUIET=true'
+				)
+
+			# disable persistent network interface names
+			if grub_settings.get('disable_pnin') is True:
+				print "grub: disabling pnin"
+				sed_i(
+					grub_def,
+					'^GRUB_CMDLINE_LINUX=""',
+					'GRUB_CMDLINE_LINUX="net.ifnames=0 biosdevname=0"'
+				)
 		else:
-			sed_i(grub_def, '^GRUB_TIMEOUT=[0-9]+', 'GRUB_TIMEOUT=0\n'
-													'GRUB_HIDDEN_TIMEOUT=0\n'
-													'GRUB_HIDDEN_TIMEOUT_QUIET=true')
-		# optionally, disable persistent network 
-		if info.manifest.system.get('grub_disable_pnin') is True:
-			# print "grub: disabling persistent network interface names"
-			sed_i(grub_def, '^GRUB_CMDLINE_LINUX_DEFAULT="quiet"',
-							'GRUB_CMDLINE_LINUX_DEFAULT="net.ifnames=0 biosdevname=0"')
-			sed_i(grub_def, '^GRUB_CMDLINE_LINUX=""',
-							'GRUB_CMDLINE_LINUX="net.ifnames=0 biosdevname=0"')
-		else:
-			sed_i(grub_def, '^GRUB_CMDLINE_LINUX_DEFAULT="quiet"',
-							'GRUB_CMDLINE_LINUX_DEFAULT="console=hvc0"')
+			sed_i(
+				grub_def,
+				'^GRUB_TIMEOUT=[0-9]+',
+				'GRUB_TIMEOUT=0\n'
+				'GRUB_HIDDEN_TIMEOUT=0\n'
+				'GRUB_HIDDEN_TIMEOUT_QUIET=true'
+			)
+
+		# quiet the console
+		sed_i(
+			grub_def,
+			'^GRUB_CMDLINE_LINUX_DEFAULT="quiet"',
+			'GRUB_CMDLINE_LINUX_DEFAULT="console=hvc0 console=ttyS0"'
+		)
 
 		# disable the recovery menu optins
 		sed_i(grub_def, '^#GRUB_DISABLE_RECOVERY="true"', 'GRUB_DISABLE_RECOVERY="true"')
@@ -77,10 +110,14 @@ class InstallGrub_1_99(Task):
 				device_map.write('(hd0) {device_path}\n'.format(device_path=device_path))
 				if not isinstance(p_map, partitionmaps.none.NoPartitions):
 					for idx, partition in enumerate(info.volume.partition_map.partitions):
-						device_map.write('(hd0,{prefix}{idx}) {device_path}\n'
-										 .format(device_path=partition.device_path,
-												 prefix=partition_prefix,
-												 idx=idx + 1))
+						device_map.write(
+							'(hd0,{prefix}{idx}) {device_path}\n'
+							.format(
+								device_path=partition.device_path,
+								prefix=partition_prefix,
+								idx=idx + 1
+							)
+						)
 
 			# Install grub
 			log_check_call(['chroot', info.root, 'grub-install', device_path])
--- a/bootstrapvz/common/tasks/grub.py
+++ b/bootstrapvz/common/tasks/grub.py
@@ -26,11 +26,6 @@ class ConfigureGrub(Task):
 		from bootstrapvz.common.tools import sed_i
 		grub_def = os.path.join(info.root, 'etc/default/grub')
 
-		print dir(info.manifest.system)
-		print info.manifest.system
-		print dir(info.manifest.system['grub'])
-		print info.manifest.system['grub']
-
 		# disable the graphical grub menu
 		sed_i(grub_def, '^#GRUB_TERMINAL=console', 'GRUB_TERMINAL=console')
 
@@ -40,7 +35,6 @@ class ConfigureGrub(Task):
 
 			# enable grub menu
 			if grub_settings.get('enable_timeout') > 0:
-				print "grub: enabling timeout"
 				sed_i(
 					grub_def,
 					'^GRUB_TIMEOUT=[0-9]+',
@@ -57,7 +51,6 @@ class ConfigureGrub(Task):
 
 			# disable persistent network interface names
 			if grub_settings.get('disable_pnin') is True:
-				print "grub: disabling pnin"
 				sed_i(
 					grub_def,
 					'^GRUB_CMDLINE_LINUX=""',
--- a/bootstrapvz/common/tasks/grub.py
+++ b/bootstrapvz/common/tasks/grub.py
@@ -26,11 +26,27 @@ class ConfigureGrub(Task):
 		from bootstrapvz.common.tools import sed_i
 		grub_def = os.path.join(info.root, 'etc/default/grub')
 		sed_i(grub_def, '^#GRUB_TERMINAL=console', 'GRUB_TERMINAL=console')
-		sed_i(grub_def, '^GRUB_CMDLINE_LINUX_DEFAULT="quiet"',
-		                'GRUB_CMDLINE_LINUX_DEFAULT="console=hvc0"')
-		sed_i(grub_def, '^GRUB_TIMEOUT=[0-9]+', 'GRUB_TIMEOUT=0\n'
-		                                        'GRUB_HIDDEN_TIMEOUT=0\n'
-		                                        'GRUB_HIDDEN_TIMEOUT_QUIET=true')
+		
+		# optionally, enable grub menu
+		if info.manifest.system.get('grub_enable_timeout') is True:
+			# print "grub: enabling menu"
+			sed_i(grub_def, '^GRUB_TIMEOUT=[0-9]+', 'GRUB_TIMEOUT=5')
+		else:
+			sed_i(grub_def, '^GRUB_TIMEOUT=[0-9]+', 'GRUB_TIMEOUT=0\n'
+													'GRUB_HIDDEN_TIMEOUT=0\n'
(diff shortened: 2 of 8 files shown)


bootstrapvz/base/manifest-schema.yml
```diff
@@ -45,6 +45,8 @@ properties:
         enum: [i386, amd64, arm64]
       userspace_architecture:
         enum: [i386]
+      grub_enable_timeout: {type: boolean}
+      grub_disable_pnin: {type: boolean}
```
myhro: Don't you think that instead of using a namespaced option, it would be better if there was a key named `grub` under `bootloader`, where you can set those options? Albeit it is not clear to me what `pnin` would mean, so a bit more verbose name would be nice.

Also, if there's an option to enable the timeout, wouldn't make sense to define the timeout value? I wouldn't expect to set it to `true` don't knowing how much this means.

bootstrapvz/common/tasks/grub.py
```diff
@@ -26,11 +26,27 @@ def run(cls, info):
...
-		                                        'GRUB_HIDDEN_TIMEOUT_QUIET=true')
+		
+		# optionally, enable grub menu
+		if info.manifest.system.get('grub_enable_timeout') is True:
+			# print "grub: enabling menu"
+			sed_i(grub_def, '^GRUB_TIMEOUT=[0-9]+', 'GRUB_TIMEOUT=5')
+		else:
+			sed_i(grub_def, '^GRUB_TIMEOUT=[0-9]+', 'GRUB_TIMEOUT=0\n'
+													'GRUB_HIDDEN_TIMEOUT=0\n'
+													'GRUB_HIDDEN_TIMEOUT_QUIET=true')
```
myhro: Looks like you aligned those lines with tab instead of spaces. Please take a look at the project's [coding style][1].

[1]: https://github.com/andsens/bootstrap-vz/blob/72699d09c8d18fccb6f434023f33c92ca4ad94ca/CONTRIBUTING.rst#coding-style

bootstrapvz/common/tasks/grub.py
```diff
@@ -26,11 +26,27 @@ def run(cls, info):
...
+		# optionally, disable persistent network 
+		if info.manifest.system.get('grub_disable_pnin') is True:
+			# print "grub: disabling persistent network interface names"
+			sed_i(grub_def, '^GRUB_CMDLINE_LINUX_DEFAULT="quiet"',
+							'GRUB_CMDLINE_LINUX_DEFAULT="net.ifnames=0 biosdevname=0"')
+			sed_i(grub_def, '^GRUB_CMDLINE_LINUX=""',
+							'GRUB_CMDLINE_LINUX="net.ifnames=0 biosdevname=0"')
```
myhro: This has to be done two times, for both `GRUB_CMDLINE_LINUX_DEFAULT` and `GRUB_CMDLINE_LINUX`?

Also, in this case, `console=hvc0` isn't needed anymore?

bootstrapvz/common/tasks/grub.py
```diff
@@ -62,9 +78,9 @@ def run(cls, info):
 				if not isinstance(p_map, partitionmaps.none.NoPartitions):
 					for idx, partition in enumerate(info.volume.partition_map.partitions):
 						device_map.write('(hd0,{prefix}{idx}) {device_path}\n'
-						                 .format(device_path=partition.device_path,
-						                         prefix=partition_prefix,
-						                         idx=idx + 1))
+										 .format(device_path=partition.device_path,
+												 prefix=partition_prefix,
+												 idx=idx + 1))
```
myhro: Again, please align with spaces instead of tabs (which are used for indentation only).

bootstrapvz/common/tools.py
```diff
@@ -112,9 +112,31 @@ def load_data(path):
...
+	"""Retrieve settings from a json or yaml config file
+	Pass it a linear array of strings, it will recursively search into a nested
+	json or yaml confg file. 
+	"""
 	config = load_data(path)
+	
+	# iterate over a linear list of strings
 	for key in config_path:
-		config = config.get(key)
+
```
myhro: This blank line at the start of block should be removed.

bootstrapvz/common/tools.py
```diff
@@ -112,9 +112,31 @@ def load_data(path):
...
+						"', in the file '" + path + "', try adding it.")
+
+		# attempt to retrieve the string
+		try:
+			config = config.get(key)
+		except AttributeError:
+			raise Exception("Error: cannot find the key '" + str(key) +
+							"', in the file '" + path + "', try adding it.")
+
+		# ensure we have a value to pass back
```
myhro: Those comments aren't needed, are they?

bootstrapvz/common/tools.py
```diff
@@ -112,9 +112,31 @@ def load_data(path):
...
+		if config is None:
+			raise Exception("Error: the key '" + str(key) +
+						" in the file '" + path + "' is not empty, try defining it.")
```
myhro: This is the case where the key `is` or `is not` empty? Looks like the condition and the message are stating opposite things.

manifests/examples/virtualbox/jessie-vagrant.yml
```diff
@@ -4,7 +4,7 @@ provider:
   name: virtualbox
   guest_additions: /usr/share/virtualbox/VBoxGuestAdditions.iso
 bootstrapper:
-  workspace: /target
+  workspace: /target/jessie
```
myhro: You don't have to create a subfolder in the workspace directory. This is [already generated in a random way][1].

[1]: https://github.com/andsens/bootstrap-vz/blob/72699d09c8d18fccb6f434023f33c92ca4ad94ca/bootstrapvz/base/bootstrapinfo.py#L19

manifests/examples/virtualbox/stretch-vagrant.yml
```diff
@@ -0,0 +1,31 @@
+---
+name: debian-{system.release}-{system.architecture}-{%y}{%m}{%d}
+provider:
+  name: virtualbox
+  guest_additions: /usr/share/virtualbox/VBoxGuestAdditions.iso
+bootstrapper:
+  workspace: /target/stretch
```
myhro: Same note about workspace here.

myhro:
Apollo,

I've made some comments on your changes. Please also take a look at the error messages regarding syntax on Travis CI.

If we are going to make some changes on the manifest, we'll also need to document them at the [appropriate section of the README][1].

[1]: https://github.com/andsens/bootstrap-vz/tree/72699d09c8d18fccb6f434023f33c92ca4ad94ca/manifests#system

manifests/examples/virtualbox/stretch-vagrant.yml
apolloclark: I added unique folder names to make it easier to run vagrant after multiple builds (jessie, stretch, sid). After each build, I'll run

```Shell
vagrant box add <build_name> <box_file>
vagrant init <build_name>
vagrant up
```

This can't be done when multiple builds are being sent to the same folder.

apolloclark:
Yup, will do. It'll take me a while to figure out how to do nesting in the manifest schema for the grub commands. Any ideas why I still need to run "grub-update" after the build for it to detect the ethernet card in Virtualbox?

apolloclark:
Ok, I updated the manifest settings, fixed the whitespace issues. Check it out now.

apolloclark:
I updated the pull request, added documentation, fixed whitespace issues. I can confirm this fix works with Jessie, and enables Stretch. However, in Virtualbox, the user still needs to run "update-grub" upon first boot. Any ideas?

apolloclark:
Well, that was easy... I just moved the "update-grub" command to happen after updating the config. Odd that it wasn't there already.

andsens:
Wow. That's quite a PR! And it looks like there is made a lot of headway in supporting stretch. Nice job!

What is the motivation for making `enable_timeout` and `disable_pnin` configurable? Those things should be taken care of by bootstrap-vz in an automated way.

On all cloud providers you never-ever need to see the grub menu (there is no way to interact with it), so `enable_timeout` should always be 0 there (for VirtualBox and kvm we can make it the default).

`disable_pnin` should probably just always be selected when `release >= stretch`, is there a reason not to?

bootstrapvz/common/tasks/grub.py
```diff
@@ -25,14 +25,52 @@ class ConfigureGrub(Task):
...
+			grub_def,
+			'^GRUB_CMDLINE_LINUX_DEFAULT="quiet"',
+			'GRUB_CMDLINE_LINUX_DEFAULT="console=hvc0 console=ttyS0"'
+		)
+
+		# disable the recovery menu optins
 		sed_i(grub_def, '^#GRUB_DISABLE_RECOVERY="true"', 'GRUB_DISABLE_RECOVERY="true"')
 
+		# commit the updates
+		log_check_call(['chroot', info.root, 'update-grub'])
```
andsens: The `ConfigureGrub` task was added in dab1dfa93031a43d183878b92caff6a7e6d68dc3. Before that, the grub2 install had no configuration, we just didn't move call when the Task was added.  So this is good - there is and was no reason to do it any other way than you are doing it now :-)

manifests/examples/virtualbox/stretch-vagrant.yml
andsens: Sorry to say, but your specific work process is not enough of a reason to justify a change like this :-/

However, you should be able to inline the `run_id` using manifest vars, so that your images are uniquely named.

bootstrapvz/common/tasks/grub.py
```diff
@@ -25,14 +25,52 @@ class ConfigureGrub(Task):
...
+				grub_def,
+				'^GRUB_CMDLINE_LINUX=""',
+				'GRUB_CMDLINE_LINUX="net.ifnames=0 biosdevname=0"'
+			)
+
+		# quiet the console
+		sed_i(
+			grub_def,
+			'^GRUB_CMDLINE_LINUX_DEFAULT="quiet"',
+			'GRUB_CMDLINE_LINUX_DEFAULT="console=hvc0 console=ttyS0"'
```
andsens: Where does the `ttyS0` come from?

bootstrapvz/common/tasks/grub.py
```diff
@@ -61,10 +99,14 @@ def run(cls, info):
...
-						                         prefix=partition_prefix,
-						                         idx=idx + 1))
+						device_map.write(
+							'(hd0,{prefix}{idx}) {device_path}\n'
+							.format(
+								device_path=partition.device_path,
+								prefix=partition_prefix,
+								idx=idx + 1
+							)
+						)
```
myhro: Looks like you are putting tabs (which should be used for indentation) where it should be spaces (used for alignment). I know this can be a pain some times (specially if your editor doesn't help), but this has been done in the entire code tree and doesn't looks like it will be changing soon.

This part of the code doesn't even had to be changed. I'd suggest you to drop this hunk.

myhro:
Besides the tabs/spaces issues, which can be also seen on other parts of `bootstrapvz/common/tasks/grub.py` and `bootstrapvz/common/tools.py`, I don't like comments which aren't really needed.

I'm not sure if you and @andsens agrees, but to me, they are bad as a missing comment in a complicated code snippet. There are places, [like this one][1] where they even looks to be saying the opposite of what the code actually does.

[1]: https://github.com/andsens/bootstrap-vz/pull/276/files#diff-19a8511ae618725c360f4068eb44c513R61

bootstrapvz/common/tasks/grub.py
apolloclark: I saw it mentioned here:
https://github.com/andsens/bootstrap-vz/pull/256
I had issues locally, when manually starting out a Virtualbox instance, outside of Vagrant, the console would appear to hang, when it was actually being rerouted. This prevented me from diagnosing issues. This changes fixes that.

bootstrapvz/common/tasks/grub.py
apolloclark: I honestly have no idea what you mean between "indentation" vs. "alignment". I've been writing code for 15 years in 5 language.

apolloclark:
Ok, I reverted the folder pathways in the manifest. Yeah, I forgot to update that grub.py comment, when I did a code push at 1:30 AM EST.

apolloclark:
@andsens yes, auto-disabling "persistent network interfaces names" / pnin for >=stretch is a great idea. However, I have no clue how to do that. I've never seen your code base before last week, don't understand the architecture, and have been blindly hacking code to get this to work.

bootstrapvz/common/tasks/grub.py
myhro: Sorry if you misunderstood me, Apollo. I never intended to say that you don't know how to program of something like that.

Please take a look at the following screenshots, where I configured my editor to show these invisible characters:

Tabs + spaces:

![][spaces]

Only tabs:

![][tabs]

Can you see the subtle differences? In this case, it was just one char wide (the tab version was shifted to the left), but I guess you can imagine what this can mean in other parts of the code.

What I'd like to make clear here is: we only use tabs if we are in a different block level of the code. The `.format()` method and its arguments, in this example, does not constitute a new level to need indentation. They should just be aligned, using spaces.

[spaces]: http://i.imgur.com/Nmw4Zr4.png
[tabs]: http://i.imgur.com/qa21IoX.png

apolloclark: Ok, I think I understand... it's a very odd convention that I've never seen. In the future, to make it easier for others to contribute code, you should publish a flake8 config file to detect this.

apolloclark: Ok, so I noticed the tox.ini file. Is this still being run by the TravisCI build? I only say it mention flake8.

myhro:
> ... when I did a code push at 1:30 AM EST.

I was indeed going to ask if you didn't slept from Friday to Saturday. :-)

> ... yes, auto-disabling "persistent network interfaces names" / pnin for >=stretch is a great idea. However, I have no clue how to do that.

You can import `stretch` from `bootstrapvz.common.releases` and compare `info.manifest.release` with it. You [can see an example with `jessie` here][1].

What Anders meant is that this should be done without user interaction (i.e. not needing to specify the new option under `grub` on the manifest).

[1]: https://github.com/andsens/bootstrap-vz/blob/72699d09c8d18fccb6f434023f33c92ca4ad94ca/bootstrapvz/common/tasks/boot.py#L37

bootstrapvz/common/tasks/grub.py
myhro: > Ok, so I noticed the tox.ini file. Is this still being run by the TravisCI build?

Yes. I can [see that the test run finished with][1] `The command "tox" exited with...`.

> Ok, I think I understand... it's a very odd convention that I've never seen.

I had also never seen it before starting to contribute with bootstrap-vz a couple years ago. :-)

Now that you understood, can you please do this on the parts of your PR that are aligned with tabs? If you don't, there's no problem, but I'll have to do this after the merge, and there will be my name on the blame.

> In the future, to make it easier for others to contribute code, you should publish a flake8 config file to detect this.

I don't know if flake8 currently supports this check, so I'll have to look onto it.

[1]: https://travis-ci.org/andsens/bootstrap-vz/builds/106044377

andsens:
> However, I have no clue how to do that. I've never seen your code base before last week, don't understand the architecture, and have been blindly hacking code to get this to work.

No problem, the `release` thing/feature is also sparsely documented, hmm, maybe I should do something about that :-)

Regardless, [check this out](https://github.com/andsens/bootstrap-vz/blob/72699d09c8d18fccb6f434023f33c92ca4ad94ca/bootstrapvz/providers/ec2/__init__.py#L76). Boiled down, what you do is:
```
from bootstrapvz.common.releases import stretch
if manifest.release > stretch:
    do_something()
```

As you can see in the link, this is done in the `resolve_tasks()` function. You *can* do it in a task directly, but this suggests that you are directing the control-flow inside the task, rather than (de-)selecting smaller tasks and then describing what is done by using the tasklist ([check out the docs](https://bootstrap-vz.readthedocs.org/en/master/developers/contributing.html#roll-complexity-into-which-tasks-are-added-to-the-tasklist) to see why it is better to not have a lot of control-flow logic inside a task).

That brings me to another point I didn't think of before. Rather than having a load of `if/else` blocks in a huge `ConfigureGrub` task you should split those blocks up into separate tasks and name them accordingly. This makes everything *a lot* more flexible, since different plugins/providers can pick and choose from those tasks at their leisure.

The best way to hide away the complexity of those smaller tasks is using [task groups](https://github.com/andsens/bootstrap-vz/blob/72699d09c8d18fccb6f434023f33c92ca4ad94ca/bootstrapvz/common/task_groups.py#L147). All providers use them, so the logic for e.g. selecting when to disable pnin is automatically reused across all of them.

apolloclark:
@andsens  it would help if you wrote some documentation, with a simple graphic chart, explaining the architecture of bootstrap-vz, and simple code examples of a basic provider, task, and plugin. I've been stumbling around blindly. It's a great project, very useful!

andsens:
> with a simple graphic chart

Good idea! I'll try firing up Inkscape and see what I can come up with.

> explaining the architecture of bootstrap-vz,

> and simple code examples of a basic provider

I think I'll just end up re-coding [the VirtualBox provider](https://github.com/andsens/bootstrap-vz/tree/72699d09c8d18fccb6f434023f33c92ca4ad94ca/bootstrapvz/providers/virtualbox) tbh. Maybe the docs could reference it as a good example (provided I add some more comments).

> task

[Is this too bare bones](https://bootstrap-vz.readthedocs.org/en/master/developers/index.html#tasks)?

> and plugin

[or this](https://bootstrap-vz.readthedocs.org/en/master/developers/plugins.html)?

apolloclark:
Anders, cool, looks good. I hadn't noticed that the first time reading through the documentation. I just changed the "disable_pnin" to auto-activate on any distro newer than Jessie, and tried to fix the whitespace issues.

bootstrapvz/common/tasks/grub.py
```diff
@@ -25,14 +25,53 @@ class ConfigureGrub(Task):
...
+		# configure the console
+		sed_i(
+		    grub_def,
+		    '^GRUB_CMDLINE_LINUX_DEFAULT="quiet"',
+		    'GRUB_CMDLINE_LINUX_DEFAULT="console=hvc0 console=ttyS0"'
+		)
+
+		# disable persistent network interface names
+		from bootstrapvz.common.releases import jessie
+		if info.manifest.release > jessie:
```
andsens: While virtually this is really the same as `info.manifest.release >= stretch`, semantically there is a difference. We specifically add something for stretch and higher up rather than doing something as long as its not jessie or lower. When reading the code you want to actually see the word `stretch`, this way you know what release some specific code concerns itself with.

apolloclark: Agreed, I updated the code.

andsens:
Hi @apolloclark. Apologies for the delay, github doesn't notify you when new commits have been added to a PR :-/

I wanted to merge this, but what's up with the dashes in front of every commit? Could you change that? (and also maybe rebase the "fixed previous commit"-commits away)

andsens:
I have implemented PNIN and credited you.
After reviewing things more thoroughly I am not to keen on merging this PR, since it contains a lot of other unrelated changes.

Closing...

apolloclark:
@andsens ok, thanks!

PR #281, opened 2016-02-10
[set localtime]: Set /etc/localtime by either copying or symlinking

srivasta:
Starting in version 2016a-1 (circa Jan 2016) the package tzdata
changed /etc/timezone into a symlink. This change is in unstable,
testing, and will be in the next release (stretch). This commit checks
the release, and creates a task to either copy the contents of the
zoneinfo file (jessie and older) or create a symbolic link (newer than
jessie).

Signed-off-by: Manoj Srivastava <srivasta@golden-gryphon.com>

### DIFF
--- a/bootstrapvz/common/task_groups.py
+++ b/bootstrapvz/common/task_groups.py
@@ -33,16 +33,7 @@ def get_standard_groups(manifest):
 	group.extend(get_network_group(manifest))
 	group.extend(get_apt_group(manifest))
 	group.extend(security_group)
-
-	from bootstrapvz.common.releases import jessie
-	if manifest.release > jessie:
-		if not locale.SetLocalTimeLink in locale_group:
-			locale_group.extend(locale.SetLocalTimeLink)
-	else:
-		if not locale.SetLocalTimeCopy in locale_group:
-			locale_group.extend(locale.SetLocalTimeCopy)
-	group.extend(locale_group)
-
+	group.extend(get_locale_group(manifest))
 	group.extend(get_bootloader_group(manifest))
 	group.extend(cleanup_group)
 	return group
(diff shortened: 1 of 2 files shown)


bootstrapvz/common/task_groups.py
```diff
@@ -33,7 +33,16 @@ def get_standard_groups(manifest):
 	group.extend(get_network_group(manifest))
 	group.extend(get_apt_group(manifest))
 	group.extend(security_group)
+
+	from bootstrapvz.common.releases import jessie
+	if manifest.release > jessie:
+		if not locale.SetLocalTimeLink in locale_group:
+			locale_group.extend(locale.SetLocalTimeLink)
```
andsens: `extend()` only takes another list as an argument. You want to use `append`.
Also, `standard_groups()` is really not meant for adding individual tasks. The tasks you got are a part of the `locale_group`, which right now is a simple list. Convert that group to a function like `get_network_group()`.

andsens:
Nice and simple PR, kudos! :-)

bootstrapvz/common/task_groups.py
srivasta: Done.

bootstrapvz/common/task_groups.py
```diff
@@ -33,7 +33,7 @@ def get_standard_groups(manifest):
 	group.extend(get_network_group(manifest))
 	group.extend(get_apt_group(manifest))
 	group.extend(security_group)
-	group.extend(locale_group)
+	group.extend(get_locale_group)
```
andsens: *`(manifest)`

srivasta: Thanks for catching that. Fixed. Also, tested.

andsens:
Looks great!. Merging.

Thank you @srivasta

Issue #282, opened 2016-02-11
Allow debconf values to be specified


nbraud:
Currently, it seems impossible to specify `debconf(7)` values for preconfiguring packages.

Should there be interest (and I assume so, given that the ability to configure package is generaly useful), I can have a go at implementing this.

andsens:
There's a fine line between configuration management and bootstrapping with preconfiguration ([as discussed previously](https://github.com/andsens/bootstrap-vz/pull/215#issuecomment-96842276)). Seeing how this is quite close to package installation itself though, I am inclined to say that it does fall within purview of bootstrap-vz. @myhro you got any opinion on this?

nbraud:
@andsens That's a pretty good point.  It made sense to me partly because my mental model of the manifest is fairly similar to a preseed file (in function, not structure).
More generally, I believe that it makes sense to be able to control that part of package installation without going as far as requiring full-blown config management  :-)

Thanks a lot for the speedy answer.

myhro:
> @myhro you got any opinion on this?

I'm not sure if I understood the issue properly. Is somewhat like [what we have][1] for the cloud-init plugin?

[1]: https://github.com/andsens/bootstrap-vz/blob/518efec1aa2a62ea65ff7dc075af47bb1e1e36a4/bootstrapvz/plugins/cloud_init/tasks.py#L60

nbraud:
@myhro I'm not sure I understand what that line does (more specifically, what `sources` is), but it's indeed about running `debconf-set-selections` in the chroot before installing packages.

myhro:
Nicolas,

> ... (more specifically, what `sources` is) ...

It is just a string, like you can see from [the previous lines][1]. E.g.: `Ec2` (with capital `e`) for the EC2 provider.

-

Anders,

I'm sitting in a wall here. In a way, it looks like something that is supposed to be done by an orchestration tool... But as you said, it is really close to package installation and we already do this on one specific case. So, I guess that the outcome is: yes, we should definitely add this to our TODO list. :-)

[1]: https://github.com/andsens/bootstrap-vz/blob/518efec1aa2a62ea65ff7dc075af47bb1e1e36a4/bootstrapvz/plugins/cloud_init/tasks.py#L52

nbraud:
Just to avoid duplicating work, I have an implementation of that.
I just need to check a few more things (including whether I need authorization from `${WORK}` to contribute patches) and I will submit a PR.

PR #290, opened 2016-02-23
Add a debconf plugin

nbraud:
This is my take on #282.

This lets you set debconf answers before extra packages are installed (but after the OS setup).

### DIFF
--- a/bootstrapvz/plugins/debconf/schema.yaml
+++ b/bootstrapvz/plugins/debconf/schema.yaml
@@ -1,0 +1,14 @@
+$schema: http://json-schema.org/schema#
+title: Manifest schema for the debconf plugin
+type: object
+properties:
+  plugins:
+    type: object
+    properties:
+      debconf:
+        name: Debconf selections to set
+        description: >-
+          This value should be an inline string in the
+          input format of debconf-set-selections(1).
+        type: string
+    required: [debconf]
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+2016-02-23
+----------
+Nicolas Braud-Santoni:
+	* #282, #290: Added 'debconf' plugin
+	* #290: Relaxed requirements on plugins manifests
+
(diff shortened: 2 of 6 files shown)


andsens:
Nice! Good catch with the `additional_properties`.
The reason plugin configs are forced to be arrays is because of forwards compatibility. If you wanted to add a setting to a plugin that is configured with a single string, you'd have to break the manifests people are already using. By making it an object you can always just add another key alongside the existing one.

Also, what about this debconf parameter:
```
       -opackage, --owner=package
           Tell debconf what package the command it is running is a part of.
           This is necessary to get ownership of registered questions right,
           and to support unregister and purge commands properly.
```

Wouldn't you need a way to specify that?

nbraud:
@andsens This plugin never calls `debconf(1)`.
The name of the package that owns a given question is always the first value on the line (see `debconf-set-selections(1)`).

Regarding forward-compatibility, I agree with the argument but I would expect package writers to choose on a case-by-case basis: here, I know that the scope of this plugin is feeding answers to `debconf(8)`, and the input format for `debconf(8)` is a plain string, so it seems like a reasonable choice.

andsens:
> would expect package writers to choose on a case-by-case basis

True. It seems it would be fairly trivial to predict whether a plugin is never-ever going to get another parameter, especially if you want to keep it simple and rather a new plugin is written to handle other features.

> so it seems like a reasonable choice.

I agree. Merging :-)

bootstrapvz/plugins/debconf/schema.yaml
```diff
@@ -0,0 +1,14 @@
...
+type: object
+properties:
+  plugins:
+    type: object
+    properties:
+      debconf:
+        name: Debconf selections to set
+        description: >-
+          This value should be an inline string in the
+          input format of debconf-set-selections(1).
```
andsens: p.s.: I like this. It would be cool to pull `name` and `description` into the documentation somehow :-)

nbraud:
Thanks a lot  :-)

PR #291, opened 2016-02-24
admin_user plugin: Fix non-EC2 behaviour

nbraud:
Currently, `admin_user` attempts to add the task `AdminUserPublicKeyEC2` if no pubkey is specified, which results in an error if the provider is not EC2.

This patch introduces an additional check, and simply sets no public key if none is defined and the provider is not EC2.

### DIFF
--- a/bootstrapvz/plugins/admin_user/__init__.py
+++ b/bootstrapvz/plugins/admin_user/__init__.py
@@ -23,7 +23,7 @@ def resolve_tasks(taskset, manifest):
 		taskset.add(tasks.AdminUserPassword)
 	if 'pubkey' in manifest.plugins['admin_user']:
 		taskset.add(tasks.AdminUserPublicKey)
-	else:
+	elif manifest.provider['name'] == 'ec2':
 		taskset.add(tasks.AdminUserPublicKeyEC2)
 
 	taskset.update([tasks.AddSudoPackage,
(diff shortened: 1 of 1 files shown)


andsens:
I see the problem, but shouldn't we fail if no authentication method is provided? Maybe check if neither a password nor a pubkey is set?

nbraud:
@andsens That's not obvious: for instance, I ran into this issue while building a VM for a lecture, where students get automatically logged-in as the user in the GUI.
There, I do not necessarily want to have a password set.

andsens:
Ah. That makes sense. Hm ok, I still think we should issue a warning à la "You have not provided any authentication method for the admin user {username}". Your use-case makes sense to support (because you can add a dozen other auth methods in any number of ways), but I would guess that 90% of the time it is a mistake in the manifest, so it's better to issue one warning too much than to little.

nbraud:
@andsens Makes sense.

nbraud:
@andsens Warnings added

andsens:
Nice, looks great. You forgot to import the logging package though, and there is a pep8 error (you can check that locally with `tox -e flake8`).

nbraud:
@andsens Fixed.

andsens:
Yup! Perfect. Merging...

Issue #293, opened 2016-02-24
bootstrap-vz should avoid bind-mounts from the host


nbraud:
After encountering #292 (and before noticing a devtmpfs was still mounted), I attempted to `rm -r` the left-over workdir: this ended up removing the vast majority of my device files from `/dev`, making the host unuseable until reboot.

This exemplifies how an issue during VM build or during rollback can negatively impact the host, and should be mitigated by not mounting `/dev`, `/dev/pts`, `/proc` and `/sys` from the host (but instead creating independent instances).

andsens:
> [..] and should be mitigated by not mounting /dev, /dev/pts, /proc and /sys from the host (but instead creating independent instances).

[We only mount-bind `/dev`](https://github.com/andsens/bootstrap-vz/blob/151dcf5a85f77c96a7439497b8f75e65461a6305/bootstrapvz/common/tasks/filesystem.py#L110).
When looking through the `debootstrap` code I see they do it a little differently,
```sh
in_target mount -t proc proc /proc
in_target mount -t sysfs sysfs /sys
in_target mount -t devfs devfs /dev
```
In bootstrap-vz, this would be:
```py
root.add_mount('proc', 'proc', ['--types', 'proc'])
root.add_mount('sysfs', 'sys', ['--types', 'sysfs'])
root.add_mount('devfs', 'dev', ['--bind'])
```

(not sure about `/dev/pts` though).


> I attempted to rm -r the left-over workdir: this ended up removing the vast majority of my device files from /dev

\*ouch, sorry to hear that.

nbraud:
@andsens I have a patchset coming that solves this.

nbraud:
PS: Mounting `procfs` and `sysfs` this way basically use the host's.
Unfortunately, I don't have a good answer for `procfs`.

nbraud:
@andsens Slightly unrelated, but do you think the “fake `/dev`” approach would be relevant for `debootstrap` too?

PR #294, opened 2016-02-24
Avoid exposing the host's special mounts

nbraud:
This should avoid the kind of problem encountered in #293.

~~Currently, this breaks `InstallGrub_1_99` and `InstallGrub_2`, which need to be fixed to call `install-grub` from outside the chroot.~~


nbraud:
Actually, I'm not sure calling `install-grub` inside the chroot is the right idea.

nbraud:
@andsens The latest comment I added rely on `nbd` having been loaded with `max_part` being 1.
Also, I wasn't able to check whether that indeed lets `install-grub` run successfully.  I will check tomorrow, in case you don't have time by then.

nbraud:
@andsens Slightly unrelated, but do you think the “fake `/dev`” approach would be relevant for `debootstrap` too?

andsens:
> @andsens Slightly unrelated, but do you think the “fake /dev” approach would be relevant for debootstrap too?

It wouldn't. Whoever coded that part was the right kind of lazy about. `debootstrap` just extracts a very special tar.gz directly into `/dev` and calls it a day. It's a file you really shouldn't try extracting anywhere else :-)

> Actually, I'm not sure calling install-grub inside the chroot is the right idea.

On of the reasons for doing this might be a leftover from grub 1.99. When run on the host, the grub helper scripts make all kinds of assumptions that you can't configure your way around. More importantly though is that you *have* to run grub inside the guest. What if the host boots from extlinux or a different version of grub?

andsens:
In the previous comment I forgot to mention how impressive this PR is. That is some really cool stuff!!
I love how it demystifies the whole `/dev` magic :-)

However, before merging, this will need some heavy testing and a lot more comments.

nbraud:
> Whoever coded that part was the right kind of lazy about it. debootstrap just extracts a very special tar.gz directly into /dev and calls it a day.

Ok, I misunderstood your comment in #293 and understood that `debootstrap` actually mounts a `devfs` there.

> this will need some heavy testing and a lot more comments.

Definitely.  That's just a first stab at it because I got incredibly annoyed by #293.
I guess it gave me incentive to fix it, though  :-)

nbraud:
@andsens I added some commits, now the grub2 install works.
Grub1.99 should work too, but I didn't check

nbraud:
@andsens Did you have the opportunity to try this?

andsens:
Hoooly crap. How much documentation did you have to sift through to get this right? Wow.
I have not had the time yet, to test this. sorry. Maybe we could pull in some help (I'll definitely test it myself as well, though)?
Paging @JamesBromberger, @myhro, @srivasta and @zmarano. Could you guys try this out, this PR has the potential to fix loads of annoying errors, could make the bootstrapping process a lot more robust to failure. It also removes unnecessary interaction with the host and could potentially provide a more "clean" system.

bootstrapvz/common/tasks/filesystem.py
```diff
@@ -100,17 +100,73 @@ def run(cls, info):
...
+                        mknod(join(dev, 'nbd%i' % i),
+                              0o660 | stat.S_IFBLK,
+                              makedev(43, nbd_max_part * i))
+                        for j in xrange(nbd_max_part):
+                                mknod(join(dev, 'nbd%ip%i' % (i,j+1)),
+                                      0o660 | stat.S_IFBLK,
+                                      makedev(43, nbd_max_part * i + j+1))
+
+                                path = '/dev/mapper/nbd%ip%i' % (i,j+1)
+                                if os.path.exists(path):
```
andsens: Under which circumstances would this path not be a link? Shouldn't we consider that an error? We expect something to be there, but it's not, which means somethings fishy.

nbraud:
> How much documentation did you have to sift through to get this right?

It wasn't quite that bad  ;-)
Also, I just woke up and realised that a PID namespace might be the right answer for `/proc`, so I added that before going back to sleep.

> I have not had the time to test this yet.

No problem, it's not very urgent.  I guess one would have to be extra unlucky to hit #293.

bootstrapvz/common/tasks/filesystem.py
nbraud: Unfortunately, I couldn't find a good reference on the internals of `device-mapper`.
Basically, I'm unaware of any case where the files in `/dev/mapper` are not symlinks (except `control`).

myhro:
@andsens this is on my TODO list. Unfortunately I've been sick (it's true what they say about Zika virus) in the past few days and the time I was able to stand on the PC was used for work-related stuff. :-(

bootstrapvz/common/tasks/filesystem.py
```diff
@@ -100,17 +100,76 @@ def run(cls, info):
...
+		os.makedirs(join(dev, 'shm'), 0o777)
+
+                # The API pseudo-devices (null, zero, full, random, urandom and tty)
+		for name, major, minor in [ ('null',   1, 3), ('zero',    1, 5), ('full', 1, 7),
+		                            ('random', 1, 8), ('urandom', 1, 9), ('tty',  5, 0) ]:
+			mknod(join(dev, name), 0o666 | stat.S_IFCHR, makedev(major, minor))
+
+                # The nbd devices
+                os.makedirs(join(dev, 'mapper'), 0o755)
+                nbd_max_part = int(info.volume._module_param('nbd', 'max_part'))
```
andsens: This breaks on EC2, since the `_module_param` part only works for QEMU volumes.

andsens:
I'm having trouble getting grub2 to install:
```
[remote] Installing grub 2
[remote] Executing: chroot /target/5fec2bf1/root grub-install /dev/xvdf
[remote] Installing for i386-pc platform.
[remote] grub-install: error: failed to get canonical path of `/dev/mapper/xvdf2'.
[remote] Command 'chroot /target/5fec2bf1/root grub-install /dev/xvdf' returned non-zero exit status 1
```

nbraud:
@andsens Could I have the manifest which makes it fail?

andsens:
Ah, yes of course. Though it isn't a manifest. It's the `gpt_grub_stable` ec2 test:
`tox -e system tests.system.ec2_ebs_hvm_tests:test_gpt_grub_stable`

Or via code:
```
#!/usr/bin/env python

from tests.system.ec2_ebs_hvm_tests import test_gpt_grub_stable
from bootstrapvz.base.main import setup_loggers

setup_loggers({'--log': '-', '--color': 'default', '--debug': True})
test_gpt_grub_stable()
```

You'll need a [`build-servers.yml`](https://bootstrap-vz.readthedocs.org/en/master/remote_bootstrapping.html#build-servers-yml) for the test to be able to run.

andsens:
Hehe, soo, that was easy:
```
for name, partition in get_partitions().items():
	mknod(join(dev, name), 0o666 | stat.S_IFBLK, makedev(int(partition['major']), int(partition['minor'])))
```

I think you can even replace the nbd stuff with this (minus the symlinks). The code simply re-creates all the block devices mentioned in `/proc/partitions`, which you can fetch with [`get_partitions()`](https://github.com/andsens/bootstrap-vz/blob/cc5435237cd4139efe85dfd2787e6b99005f9c5a/bootstrapvz/common/fs/__init__.py#L4).

nbraud:
@andsens Thanks for having a look, I was unable to make the time those last weeks.

I will see about using your snippet and testing.

andsens:
OK, so I have tried getting this to work, but there are a few problems:
* After isolating the process it seems python can't start new threads. It fails with `thread.error: can't start new thread` in `/usr/lib/python2.7/threading.py:745`
* Some tasks use [`unmount()`](https://github.com/andsens/bootstrap-vz/blob/8eb51524f3bd140b7883c44924948499f0cab8f2/bootstrapvz/common/fs/__init__.py#L21) (e.g. the prebootstrapped plugin) to do things to the volume. After remounting `/dev` is of course not populated any longer, so we need to either re-create it or create `/dev` outside `root` and `mount --bind` it into `root`

I have improved on your method by simply walking through `/dev` and re-creating everything in it, you can check it out [here](https://github.com/andsens/bootstrap-vz/blob/ebd571f9cf9d56bc8d6222ed5801e0501caef7d5/bootstrapvz/common/tasks/filesystem.py#L102).
It's on the `special-devices` branch.

I am closing this PR, let me know if you find a solution to these problems. I'd love to get this working, but there is just too much other stuff for me to work on.

nbraud:
@andsens Duplicating the entirety of `/dev` still comes with the risk that the build/installation of the guest might clobber something in the host (for instance during `grub-install`).

I also didn't have time lately to dedicate to this PR, but when I do I will try sending you separate PRs.
I realise that trying to work simultaneously on `/dev`, `/proc` and `/sys` here didn't help making it very manageable.

PR #296, opened 2016-02-27
Cleanup package list and add in other packages

zmarano:
These packages were being added outside of bootstrap-vz by our build scripts.
Remove i386 kernels that GCE does not support as well as squeeze (never supported), and add in a stretch kernel.


andsens:
Nice. Wouldn't it be better to only include the bare essentials by code and the rest via the manifest? Right now you are forcing people to include all those packages, when they build their own images. If you add them via the manifest, they pick and choose themselves.

zmarano:
Sure that works too, done! Thanks for the suggestion!

andsens:
Great stuff! Merging.

p.s.: Since you don't support squeeze, you should probably disallow it using either the validation function or the manifest schema.

PR #298, opened 2016-02-29
file_copy: Fix manifest and document

nbraud:
The manifest for the `file_copy` plugin was failing to validate the data properly.

### DIFF
--- a/bootstrapvz/plugins/file_copy/README.rst
+++ b/bootstrapvz/plugins/file_copy/README.rst
@@ -1,0 +1,26 @@
+File copy
+---------
+
+This plugin lets you copy files from the host to the VM under construction,
+  create directories, and set permissions and ownership.
+
+Note that this necessarily violates the `first development guideline`_.
+
+.. _first development guideline: https://github.com/andsens/bootstrap-vz/blob/master/CONTRIBUTING.rst#the-manifest-should-always-fully-describe-the-resulting-image
+
+
+Settings
+~~~~~~~~
+
+The ``file_copy`` plugin takes a (non-empty) ``files`` list, and optionnaly a ``mkdirs`` list.
+
+Files (items in the ``files`` list) must be objects with the following properties:
+* ``src`` and ``dst`` (required) are the source and destination paths.
+  ``src`` is relative to the current directory, whereas ``dst`` is a path in the VM.
+* ``permissions`` (optional) is a permission string in a format appropriate for ``chmod(1)``.
+* ``owner`` and ``group`` (optional) are respectively a user and group specification,
+  in a format appropriate for ``chown(1)`` and ``chgrp(1)``.
+
+Folders (items in the ``mkdirs`` list) must be objects with the following properties:
+* ``dir`` (required) is the path of the directory.
+* ``permissions``, ``owner`` and ``group`` are the same as for files.
--- a/bootstrapvz/plugins/file_copy/__init__.py
+++ b/bootstrapvz/plugins/file_copy/__init__.py
@@ -3,13 +3,18 @@ import tasks
 
 def validate_manifest(data, validator, error):
 	import os.path
+
 	schema_path = os.path.normpath(os.path.join(os.path.dirname(__file__), 'manifest-schema.yml'))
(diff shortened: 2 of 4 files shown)


andsens:
Could you elaborate? It's really hard to read the diff. Was it the indentation of the `required` property?

nbraud:
@andsens It seems I didn't commit everything the first time...

Basically:
- `#/definitions/absolute_path` is not known to the validator library, resulting in `jsonschema.exceptions.RefResolutionError` exceptions.
- `files` and `mkdirs` were not required to be of type object.
- Properties were not described in a `properties` block.
- `additionalProperties: false` was not used were necessary.

I also added a couple of commits (documentation & pushing the validation into `validate_schema` rather than in a task).

nbraud:
PS: I wonder if there is a metaschema against which we could systematically validate schemata.

bootstrapvz/plugins/file_copy/README.rst
```diff
@@ -0,0 +1,26 @@
...
+
+Note that this necessarily violates the `first development guideline`_.
+
+.. _first development guideline: https://github.com/andsens/bootstrap-vz/blob/master/CONTRIBUTING.rst#the-manifest-should-always-fully-describe-the-resulting-image
+
+
+Settings
+~~~~~~~~
+
+The ``debconf`` plugin takes a (non-empty) ``files`` list, and optionnaly a ``mkdirs`` list.
```
andsens: `debconf`?? Methinks you made a copypasta mistake :-)

bootstrapvz/plugins/file_copy/README.rst
```diff
@@ -0,0 +1,26 @@
+File copy
+---------
+
+This plugin lets you copy files from the host to the VM under construction,
+  create directories, and set permissions and ownership.
+
+Note that this necessarily violates the `first development guideline`_.
```
andsens: Good catch! It does.  
Though this plugin is more for users than developers. e.g. I'd never accept PRs for example/official manifests that make use of this plugin. But it's a great way for people to pre-install a lot of stuff on their custom images (the same could also be said for the `command` and the various config management plugins).

bootstrapvz/plugins/file_copy/__init__.py
```diff
@@ -3,13 +3,19 @@
...
+	from bootstrapvz.common.exceptions import TaskError
+
 	schema_path = os.path.normpath(os.path.join(os.path.dirname(__file__), 'manifest-schema.yml'))
 	validator(data, schema_path)
 
+	for file_entry in data['plugins']['file_copy']['files']:
+		srcfile = file_entry['src']
+		if not os.path.isfile(srcfile):
+			msg = 'The source file %s does not exist.' % srcfile
+			raise TaskError(msg)
```
andsens: Use the `error` function provided as an argument instead. [Like so](https://github.com/andsens/bootstrap-vz/blob/e961438904fdbe7c88e4c8d20cc211c5026e5a5c/bootstrapvz/providers/ec2/__init__.py#L32)

andsens:
Whoa! Great work! This PR really improves the quality of the plugin a lot, thank you :-)

bootstrapvz/plugins/file_copy/README.rst
nbraud: Indeed

nbraud:
@andsens Updated.

andsens:
Nice. Merging. Thank you @nbraud!

andsens:
Woops, hold on. We got a test failure: `bootstrapvz/plugins/file_copy/__init__.py:6:2: F401 'TaskError' imported but unused`

nbraud:
I managed to test the wrong branch... twice!

Really sorry about that.  :V

bootstrapvz/plugins/file_copy/README.rst
nbraud: Yes, it's indeed pretty convenient.
I put that part in to make things more understandable for prospective contributors, but it can be removed.

andsens:
Don't worry about it, the end result is great. Nice work!

PR #299, opened 2016-02-29
manifest: Validate schemata before use

nbraud:
This modifies `manifest.schema_validator` (which is passed as argument to all the `validate_schema` functions) to:
- first validate the schema against the JSON metaschema;
- then validate the manifest against the schema.

This is probably suboptimal, as the metaschema is reloaded every time, but I noticed no significant impact.

### DIFF
--- a/bootstrapvz/base/manifest.py
+++ b/bootstrapvz/base/manifest.py
@@ -27,6 +27,11 @@ class Manifest(object):
 		if path is None and data is None:
 			raise ManifestError('`path\' or `data\' must be provided')
 		self.path = path
+
+		import os.path
+		self.metaschema = load_data(os.path.normpath(os.path.join(os.path.dirname(__file__),
+		                                                          'metaschema.json')))
+
 		self.load_data(data)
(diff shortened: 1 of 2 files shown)


andsens:
Great idea! Regarding the reloading, you could just load the metaschema in `__init__`.

nbraud:
I wasn't sure it was worth doing, but on the other hand it is basically a 2 line patch.
¯\_(ツ)_/¯

andsens:
Nice. If you could fix the test failures, I'll merge
```
bootstrapvz/base/manifest.py:31:31: F821 undefined name 'os'
bootstrapvz/base/manifest.py:31:48: F821 undefined name 'os'
bootstrapvz/base/manifest.py:31:61: F821 undefined name 'os'
bootstrapvz/base/manifest.py:126:3: F401 'os' imported but unused
```

nbraud:
Looks like I managed to test in the wrong branch...

andsens:
Merging...

PR #300, opened 2016-03-01
Travis: use the container-based test environment

nbraud:
This should help getting the builds started much faster.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: False
 language: python
 python:
     - "2.7"


nbraud:
Don't merge yet, I would like to see if this works

andsens:
> Of course, adding tests that actually exercise the entirety of bootstrap-vz will require switching back to sudo: True.

Lets cross that bridge when we get to it :-)
For now this should help quite a bit, thank you @nbraud.

PR #301, opened 2016-03-03
Fix expand-root so it works correctly on GCE and other providers

zmarano:


bootstrapvz/providers/ec2/__init__.py
```diff
@@ -89,6 +89,7 @@ def resolve_tasks(taskset, manifest):
 
 	if manifest.volume['partitions']['type'] != 'none':
 		taskset.add(initd.AdjustExpandRootScript)
+		taskset.add(initd.AdjustGrowpartWorkaround)
```
andsens: Since this is now its own task, you should do the `>=jessie` check here. Otherwise the task is a no-op for earlier distributions.

zmarano: Good point. Fixed.

andsens:
Looks great! Merging...
p.s.: Do we need it for anything higher than jessie? Doesn't cloud-init growpart on stretch etc. run fine?

zmarano:
I actually do know the history of the problems here. Basically the version of cloud-utils (which contains growpart) in jessie has a bug when expanding MBR disks >2TB and will start overwriting the disk and you can end up with a bricked disk (for instance, if you made a 2.1TB disk you would end up with a 0.1TB partition). That bug was fixed long before Jessie shipped- and despite an open bug on the subject- was never fixed before Jessie's release. Very recently cloud-utils was updated and pulled into jessie-backports. It should therefore also be fixed in stretch. I have not however had time to look at this and verify if this is the case or if other bugs may have been introduced with that patch. For GCE, we built a hack around this to just do nothing when the disk is >2TB- but I'm planning on removing that in favor of the patched version from backports (when I get time to verify and test it out).

I don't know if the EC2 image in question here has other intentions and/or other changes in their growpart script they pulled into their assets or not. I don't want to comment for their intentions.

PR #303, opened 2016-03-03
Again fix the expand-root script for GCE.

zmarano:
The volume is not /dev/loop0 but it should be /dev/sda.

Outside the scope of fixing my immediate problem here, should there be something in the manifest for volume to denote the resulting image's device_path instead of the install device_path (which will change depending upon what kind of volume you are using). If you agree, I'll create an issue for that because it seems like something you should be able to specify in a manifest file.

### DIFF
--- a/bootstrapvz/providers/gce/__init__.py
+++ b/bootstrapvz/providers/gce/__init__.py
@@ -40,6 +40,7 @@ def resolve_tasks(taskset, manifest):
 	                tasks.host.InstallHostnameHook,
 	                tasks.boot.ConfigureGrub,
 	                initd.AddExpandRoot,
+	                tasks.initd.AdjustExpandRootDev,
 	                initd.InstallInitScripts,
 	                boot.BlackListModules,
 	                boot.UpdateInitramfs,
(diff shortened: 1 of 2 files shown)


andsens:
> The volume is not /dev/loop0 but it should be /dev/sda.

Oh, crap. This means right now it also breaks for ec2, docker and all the others?

> Outside the scope of fixing my immediate problem here, should there be something in the manifest for volume to denote the resulting image's device_path instead of the install device_path [?]

No, definitely not. This is something bootstrap-vz is supposed to figure out for the user. If push comes to shove, we can even do it by uuid. We shouldn't expect the user to know what the name of the root device is going to be.

zmarano:
It "may" break for ec2 and docker yes. The behavior for non loop devices is probably going to be different. I do suspect it will break but don't have a great way to verify that. Do you want me to create an issue?

andsens:
That'd nice, yes please :-)
Then we could verify and track where it fails/does not fail.

I'm heading to bed now (it's 2 am here).

zmarano:
Done: https://github.com/andsens/bootstrap-vz/issues/304

Sleep well!

PR #305, opened 2016-03-11
Enable setting `apt.conf(5)` options

KellerFuchs:
This adds the possibility of adding configuration snippets in `apt.conf` and `apt.conf.d`:
```yaml
packages:
    apt.conf.d:
        00InstallRecommends:  >-
            APT::Install-Recommends "false";
            APT::Install-Suggests   "false";
```

Those settings are applied before packages get installed, making it viable to set things like `APT::Install-Recommends` or `Acquire::http::Proxy`.

Moreover, 2 fixup commits are included.

### DIFF
--- a/bootstrapvz/base/__init__.py
+++ b/bootstrapvz/base/__init__.py
@@ -32,3 +32,14 @@ def validate_manifest(data, validator, error):
 
 		if release == squeeze:
 			error('Grub installation on squeeze is not supported', ['system', 'bootloader'])
+
+        # Check the provided apt.conf(5) options
+        if 'packages' in data:
+                for name, val in data['packages'].get('apt.conf.d', {}).iteritems():
+                        from bootstrapvz.common.tools import log_call
+
+                        status, _, _ = log_call(['apt-config', '-c=/dev/stdin', 'dump'],
+                                                stdin=val + '\n')
+
+                        if status != 0:
+                                error('apt.conf(5) syntax error', ['packages', 'apt.conf.d', name])
--- a/bootstrapvz/base/__init__.py
+++ b/bootstrapvz/base/__init__.py
@@ -6,40 +6,40 @@ __all__ = ['Phase', 'Task', 'main']
 
 
 def validate_manifest(data, validator, error):
-	"""Validates the manifest using the base manifest
+    """Validates the manifest using the base manifest
 
-	:param dict data: The data of the manifest
-	:param function validator: The function that validates the manifest given the data and a path
-	:param function error: The function tha raises an error when the validation fails
-	"""
-	import os.path
-	schema_path = os.path.normpath(os.path.join(os.path.dirname(__file__), 'manifest-schema.yml'))
-	validator(data, schema_path)
+    :param dict data: The data of the manifest
+    :param function validator: The function that validates the manifest given the data and a path
+    :param function error: The function tha raises an error when the validation fails
+    """
+    import os.path
+    schema_path = os.path.normpath(os.path.join(os.path.dirname(__file__), 'manifest-schema.yml'))
+    validator(data, schema_path)
 
-	from bootstrapvz.common.releases import get_release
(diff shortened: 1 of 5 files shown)


KellerFuchs:
Do not merge yet, this is untested.

andsens:
Hi @KellerFuchs. Quoting the dev guidelines:
> Please try to be very descriptive about your changes when you write a pull request, stating what it does, **why it is needed**, **which use cases this change covers**, etc

Could you elaborate on the use case?
Also, maybe you could add an example to the docs on how to use the setting.

p.s.: Thanks for the fixes in the documentation.

KellerFuchs:
@andsens Thanks for the feedback.  I added a quick description here, and an example in the example manifest in the docs.

KellerFuchs:
@andsens Is there anything else missing in that PR?

andsens:
Nope. I'll try it out tomorrow and merge if all's well :-)

KellerFuchs:
@andsens Thanks a lot for reviewing.

andsens:
My apologies @KellerFuchs, testing #294 takes quite som time and I try to treat PRs in a FIFO manner, so it's going to be a while before I can test and then merge this (unless @myhro can help?)

myhro:
Hi Anders,

> unless @myhro can help?

Yes I can, albeit this pull request is so carefully crafted that I have almost nothing to ask for a fix. The only part that I didn't quite understood was the change from `minItems` to `minProperties` the base manifest schema.

@KellerFuchs can you please take care of the following items?

* Fix the "8-spaces indentation". We are still asking for "[indent with tabs and align with spaces][1]" in the coding style guidelines, but I've talked with Anders about moving from tabs to 4-spaces indentation in the near future and a few PRs have been accepted this way already. So, it's OK if you go from 8 to 4 spaces indentation instead of tabs right now.
* Rebase it on top of `master`. Right now it is 3 commits behind.
* The last one is not a must, but would be nice if we change the example of `10proxy`. I like the idea of having another entry as an example of a single-line file (which doesn't needs `>-` in its syntax), but we already have the [apt_proxy plugin][2]. It has an option to make the configuration persistent (to be used only during the bootstrap process or remain on the resulting image), can check if the server can be reached, has a nice syntax for `username`/`password`, etc. This way it will not look like we are encouraging the usage of a hand-made setting instead of a plugin designed specifically for this task.

Thanks for your great work!

Regards,
Tiago.

[1]: https://github.com/andsens/bootstrap-vz/blob/1b8749bd7fd36312918f9a4c68c39336a5a07971/CONTRIBUTING.rst#coding-style
[2]: https://github.com/andsens/bootstrap-vz/tree/1b8749bd7fd36312918f9a4c68c39336a5a07971/bootstrapvz/plugins/apt_proxy

KellerFuchs:
@myhro As far as I understand, `minItems` is only good for arrays, [not for objects](https://spacetelescope.github.io/understanding-json-schema/reference/object.html#size).
I will rebase and push fixes right now.

KellerFuchs:
@myhro Done.

myhro:
Many thanks, @KellerFuchs. As some files had their indentation changed in its entirety, I have to ask @andsens: do you see a problem with this? Or can we merge the PR in the current state?

I'm in favor for the later option, as we are going to change all of them anyway.

andsens:
Well initially I'd say that we should preserve indentation until the change so that the changesets are readable, but honestly, the change to spaces indentation is going to screw that over anyways. You'll be able to blame all files and see my name on them, heh :-/

So this is fine, go ahead and merge.

myhro:
> You'll be able to blame all files and see my name on them, heh :-/

Yeah, this is something that we can't get rid of... Thank you, Anders.

Thanks again, @KellerFuchs.

Merging...

KellerFuchs:
@andsens @myhro Thanks a lot for reviewing.

PR #306, opened 2016-03-15
fix document toctree

hitsumabushi:
Fix invalid item levels.
All manifest settings should be same toctree level like following.

- EC2
  - Manifest settings
- Google Compute Engine
  - Manifest settings
- KVM
  - Manifest settings


myhro:
Hi @hitsumabushi ,

Nice catch! Although the provider name was indeed rendered as `h1`, the `Manifest settings` title wasn't being rendered as `h2`.

Thank you.
Merging...

PR #309, opened 2016-03-29
Add busybox as a required package for GCE.

zmarano:
busybox as a dependency seems to have disappeared at some point in recent PR's and is required for initrd with the growroot disable script that currently exists in GCE images.

### DIFF
--- a/bootstrapvz/providers/gce/tasks/packages.py
+++ b/bootstrapvz/providers/gce/tasks/packages.py
@@ -15,6 +15,7 @@ class DefaultPackages(Task):
 	@classmethod
 	def run(cls, info):
 		info.packages.add('acpi-support-base')
+		info.packages.add('busybox')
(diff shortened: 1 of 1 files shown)


andsens:
I checked the history. Actually, it never existed, maybe it was a dependency for some other package that we aren't pulling in any longer?

Merging.

zmarano:
Yeah it was not something we explicitly depended on. I was trying to find what other dependencies would have changed elsewhere with no luck however. In any case, adding it explicitly works for us.

zmarano:
Oops- I hit the wrong button and closed this by accident, can you merge this PR?

andsens:
Heh, I said I was merging, but apparently I forgot to press the button. Pressing it now :-)

zmarano:
Thanks :)

PR #314, opened 2016-04-15
cloud_init disable AdjustGrowpartWorkaround

thepwagner:
AdjustGrowpartWorkaround modifies /etc/init.d/expand-root, which doesn't
exist if AddExpandRoot doesn't run.


andsens:
Nice catch. Merging.

PR #315, opened 2016-04-16
cloud-init: modify groups

thepwagner:
Very similar to user customization.

Use case is:
```
...
packages:
  sources:
    docker:
      - deb http://apt.dockerproject.org/repo/ debian-jessie main
  install:
    - docker-engine
plugins:
  cloud_init:
    metadata_sources: Ec2
    username: admin
    groups:
      - docker
```

`docker_daemon` requires Grub due to `EnableMemoryCgroup` and I prefer installing docker from .deb.
`commands` can't `usermod` because the user doesn't exist yet.
I ended up here.

### DIFF
--- a/bootstrapvz/plugins/cloud_init/README.rst
+++ b/bootstrapvz/plugins/cloud_init/README.rst
@@ -13,6 +13,9 @@ Settings
 
 -  ``username``: The username of the account to create.
    ``required``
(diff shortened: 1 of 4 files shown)


andsens:
Yup, seems reasonable. Merging...

PR #316, opened 2016-04-17
Fixes for uploading a GCE image

justinsb:
GCE image upload simply didn't work.  With this PR, it does.  Changes:

* syntax is `gcloud compute images`, not `gcloud compute image`
* --source-uri=<uri> should be a single argument
* Expand the description template before passing it to gcloud
* Rename image in tar file to disk.raw
* Name the image as specified in the manifest
* Documentation improvements


andsens:
Hi @justinsb, I apologize for the way-too-late answer. This looks like a good PR, I'd like to hear from one of the guys at google before I merge this though. @zmarano, can you sign off on this?

zmarano:
Looks good to me- oddly enough we didn't use this functionality in our current build process- and yes it was very broken.

andsens:
Grrreat! Merging.
Thank you both :-)

PR #317, opened 2016-04-18
Disable SSH root login on GCE images.

zmarano:
We thought this was already set but it seems to have changed at some point in history (functionally in Debian configs of boostrap-vz).
https://cloud.google.com/compute/docs/instances/connecting-to-instance#root

### DIFF
--- a/bootstrapvz/providers/gce/__init__.py
+++ b/bootstrapvz/providers/gce/__init__.py
@@ -46,6 +46,7 @@ def resolve_tasks(taskset, manifest):
 	                boot.UpdateInitramfs,
 	                ssh.AddSSHKeyGeneration,
 	                ssh.DisableSSHPasswordAuthentication,
+	                ssh.DisableRootLogin,
 	                tasks.apt.CleanGoogleRepositoriesAndKeys,
 
 	                image.MoveImage,


andsens:
Actually jessie should have this disabled per default. Also, why aren't you using the `admin_user` plugin in the manifests? How exactly do you create the admin user right now?

zmarano:
We don't include an admin user on GCE. We use user controlled SSH keys via metadata (either at the project level of instance level) and add a user to the VM with that key and add that user to sudoers (at the moment). There is no default admin account/password/key on our platform for any distro. Let me know if this doesn't make sense- it is the way GCE has always worked and is a major feature of the platform.

Jessie defaults to: PermitRootLogin without-password

andsens:
Ah. OK, that makes perfect sense then. Merging.

> Jessie defaults to: PermitRootLogin without-password

WTH? I was convinced root login was entirely disabled. Better go fix the admin plugin I guess.

zmarano:
Well, this is what I surmise is the default because we did not change it and that is what our images are currently set to. Unless something else in bootstrap-vz is changing it... you probably want to check this to be sure :)

myhro:
Can confirm. I've checked a server that was recently assembled (bare metal, installed from ISO) and indeed, it has `PermitRootLogin without-password` in its `/etc/ssh/sshd_config`.

Issue #318, opened 2016-04-27
No console with recent KVM builds


andrewbogott:
My new kvm images don't produce a visible console at all.

I've been chasing this for a while.  I suspect that it's related to the console= change in 608de63d3ed4a365d6409007805d73d1f2ea909a but I don't have a smoking gun yet.

andrewbogott:
Most likely this is related to https://github.com/andsens/bootstrap-vz/issues/312 but the associated fix doesn't seem to work for me.

andrewbogott:
Here's a more complete statement of the problem:

I'm building on Debian GNU/Linux 8.4 using a minimal (albeit custom) manifest.  When the build finishes, I test the image like this:

 sudo qemu-system-x86_64 --nographic ./debian-jessie.raw

With the current bootstrap-vz tip code, that command sits with no feedback.  I've waited as much as 10-15 minutes to no effect.  With the attached patch, the console log starts streaming by after a few seconds, eventually ending with a login prompt.

I've also tested these images by loading into an OpenStack install using kvm.  The results are similar -- Nova is unable to provide me with a console log for default builds, but with the attached patch logging appears as I would expect.

Issue #312, opened 2016-04-14
kernel boot option console=hvc0 specific to xen and not suitable for virtualbox ?


olberger:
Hi.

It seems https://github.com/andsens/bootstrap-vz/commit/608de63d3ed4a365d6409007805d73d1f2ea909a changed the default kernel boot option in order to redirect login to the xen console (?).

However, as discussed in https://lists.debian.org/debian-cloud/2016/04/msg00021.html I think it is the cause of booting issues with virtualbox+vagrant.

I'd suggest to change this to depend on the target platform

myhro:
@olberger can please you confirm if the #321 merge fixed this issue as well?

olberger:
I probably won't be able to test this very soon, but will complain for sure when I do if things aren't fixed. Sorry for not being able to help more right now.

EmmanuelKasper:
Yes this fixes the issue, at the cost of losing standard boot ouput on the normal console, since you redirect it to a serial console.

GRUB_CMDLINE_LINUX_DEFAULT="console=tty0 console=ttyS0" would be better so you have boot messages on both VGA console ans serial 
for the record, this is what vmdeboostrap is doing.

andsens:
Fixed :-)

PR #320, opened 2016-05-09
console=ttyS0 for kvm and virtualbox

andrewbogott:
This definitely fixes my issue, 318, and I suspect that it fixes 312 as well but I haven't tested.  I have my doubts about this being the right solution... the majority of providers now hack in ttys0, which makes me think that the original change to grub (608de63d3ed4a365d6409007805d73d1f2ea909a) may be in error.

An alternative to this change is to standardize on ttys0 in common/tasks/grub.py and use hvc0 only where needed.  To take that route, though, I'd need to know why the original move to hvc0 was made, so I'll leave that judgement up to Anders.

Fixes: https://github.com/andsens/bootstrap-vz/issues/318
Fixes: https://github.com/andsens/bootstrap-vz/issues/312

### DIFF
--- a/bootstrapvz/providers/kvm/__init__.py
+++ b/bootstrapvz/providers/kvm/__init__.py
@@ -1,5 +1,6 @@
 from bootstrapvz.common import task_groups
 import tasks.packages
+import tasks.boot
 from bootstrapvz.common.tasks import image
 from bootstrapvz.common.tasks import loopback
 from bootstrapvz.common.tasks import initd
@@ -23,6 +24,7 @@ def resolve_tasks(taskset, manifest):
 	                ssh.ShredHostkeys,
 	                ssh.AddSSHKeyGeneration,
 	                image.MoveImage,
+	                tasks.boot.ConfigureGrub,
 	                ])
 
 	if manifest.provider.get('virtio', []):
--- a/bootstrapvz/providers/kvm/tasks/boot.py
+++ b/bootstrapvz/providers/kvm/tasks/boot.py
@@ -1,0 +1,18 @@
+from bootstrapvz.base import Task
+from bootstrapvz.common import phases
+from bootstrapvz.common.tasks import grub
+import os
+
+
+class ConfigureGrub(Task):
+	description = 'Change grub configuration to allow for ttyS0 output'
+	phase = phases.system_modification
+	predecessors = [grub.ConfigureGrub]
+	successors = [grub.InstallGrub_1_99, grub.InstallGrub_2]
+
+	@classmethod
+	def run(cls, info):
+		from bootstrapvz.common.tools import sed_i
+		grub_def = os.path.join(info.root, 'etc/default/grub')
+		sed_i(grub_def, '^GRUB_CMDLINE_LINUX_DEFAULT="console=hvc0"',
+		                'GRUB_CMDLINE_LINUX_DEFAULT="console=ttyS0"')
--- a/bootstrapvz/providers/virtualbox/__init__.py
+++ b/bootstrapvz/providers/virtualbox/__init__.py
@@ -1,5 +1,6 @@
 from bootstrapvz.common import task_groups
 import tasks.packages
+import tasks.boot
 from bootstrapvz.common.tasks import image
 from bootstrapvz.common.tasks import loopback
 
@@ -17,6 +18,7 @@ def resolve_tasks(taskset, manifest):
 	                loopback.AddRequiredCommands,
 	                loopback.Create,
 	                image.MoveImage,
+	                tasks.boot.ConfigureGrub,
 	                ])
 
 	if manifest.provider.get('guest_additions', False):
--- a/bootstrapvz/providers/virtualbox/tasks/boot.py
+++ b/bootstrapvz/providers/virtualbox/tasks/boot.py
@@ -1,0 +1,18 @@
+from bootstrapvz.base import Task
+from bootstrapvz.common import phases
+from bootstrapvz.common.tasks import grub
+import os
+
+
+class ConfigureGrub(Task):
+	description = 'Change grub configuration to allow for ttyS0 output'
+	phase = phases.system_modification
+	predecessors = [grub.ConfigureGrub]
+	successors = [grub.InstallGrub_1_99, grub.InstallGrub_2]
+
+	@classmethod
+	def run(cls, info):
+		from bootstrapvz.common.tools import sed_i
+		grub_def = os.path.join(info.root, 'etc/default/grub')
+		sed_i(grub_def, '^GRUB_CMDLINE_LINUX_DEFAULT="console=hvc0"',
+		                'GRUB_CMDLINE_LINUX_DEFAULT="console=ttyS0"')


myhro:
Andrew,

Can you please help me to understand what is exactly the problem? [As I said][1] when [Olivier Berger originally reported his problem with VirtualBox][2], I wasn't able to reproduce it back then, and I'm still not able to do it now when looks like KVM is suffering from the same issue. Please state what system are you using to build (and run) the image, if you are using the last version of bootstrap-vz from the `master` tree, attach the manifest you can, etc.

Here's what I did trying to reproduce the issue:

* Host machine: Debian `jessie`, kernel `3.16.0-4-amd64` and `libvirt-bin` `1.2.9-9+deb8u2`.
* Built the [`jessie-virtio.yml`][3] manifest.
* Imported the resulting image (using `virt-manager`).
* Booted the machine with no problem, was able to see the GRUB menu, etc.

To make sure I've logged into the newly created machine, changed `hvc0` to `ttyS0` in `/etc/default/grub` (which is what your PR is doing), run `update-grub` and... The GRUB menu was gone! All I could see was the lines:

     Loading Linux 3.16.0-4-amd64 ...
     Loading initial ramdisk ...

Which is a little bit like what happened to Olivier, except that the boot didn't hanged and I was able to see the login prompt again. In the end, what I'm seeing looks to be the opposite from what you reported.

About your PR:

> This definitely fixes my issue, 318, and I suspect that it fixes 312 as well but I haven't tested.

Even if it did solved your issue (which I wasn't able to reproduce), please don't close issues in the commit message that you are not sure that will be solved as well.

> An alternative to this change is to standardize on ttys0 in common/tasks/grub.py and use hvc0 only where needed. 

... Or the other way around (standardized `hvc0`, use `ttyS0` when needed), which I agree.

Anyway, you shouldn't use two identical files for different providers. The `ConfigureGrub` task from both `bootstrapvz/providers/kvm/tasks/boot.py` and `bootstrapvz/providers/virtualbox/tasks/boot.py` could be perfectly moved to `bootstrapvz/common/tasks/boot.py` (and imported/added to the taskset in their `__init__.py` when needed, as you are aware of).

Regards,
Tiago.

[1]: https://lists.debian.org/debian-cloud/2016/04/msg00017.html
[2]: https://lists.debian.org/debian-cloud/2016/04/msg00016.html
[3]: https://github.com/andsens/bootstrap-vz/blob/bc1d546/manifests/examples/kvm/jessie-virtio.yml

andrewbogott:
On 5/9/16 8:42 PM, Tiago Ilieve wrote:
>
> Andrew,
>
> Can you please help me to understand what is exactly the problem? As I 
> said <https://lists.debian.org/debian-cloud/2016/04/msg00017.html> 
> when Olivier Berger originally reported his problem with VirtualBox 
> <https://lists.debian.org/debian-cloud/2016/04/msg00016.html>, I 
> wasn't able to reproduce it back then, and I'm still not able to do it 
> now when looks like KVM is suffering from the same issue. Please state 
> what system are you using to build (and run) the image, if you are 
> using the last version of bootstrap-vz from the |master| tree, attach 
> the manifest you can, etc.
>
I've updated https://github.com/andsens/bootstrap-vz/issues/318 with a 
wordier statement of my problem.  I'm attaching the full version of the 
manifest that produces the issue.

> Here's what I did trying to reproduce the issue:
>
>   * Host machine: Debian |jessie|, kernel |3.16.0-4-amd64| and
>     |libvirt-bin| |1.2.9-9+deb8u2|.
>   * Built the |jessie-virtio.yml|
>     <https://github.com/andsens/bootstrap-vz/blob/bc1d546/manifests/examples/kvm/jessie-virtio.yml>
>     manifest.
>
I would be happy to try and reproduce the bug with that manifest if you 
send it my way.  I don't see it in my git clone.

>   * Imported the resulting image (using |virt-manager|).
>   * Booted the machine with no problem, was able to see the GRUB menu,
>     etc.
>
> To make sure I've logged into the newly created machine, changed 
> |hvc0| to |ttyS0| in |/etc/default/grub| (which is what your PR is 
> doing), run |update-grub| and... The GRUB menu was gone! All I could 
> see was the lines:
>
> |Loading Linux 3.16.0-4-amd64 ... Loading initial ramdisk ... |
>
> Which is a little bit like what happened to Olivier, except that the 
> boot didn't hanged and I was able to see the login prompt again. In 
> the end, what I'm seeing looks to be the opposite from what you reported.
>
> About your PR:
>
>     This definitely fixes my issue, 318, and I suspect that it fixes
>     312 as well but I haven't tested.
>
> Even if it did solved your issue (which I wasn't able to reproduce), 
> please don't close issues in the commit message that you are not sure 
> that will be solved as well.
>
>     An alternative to this change is to standardize on ttys0 in
>     common/tasks/grub.py and use hvc0 only where needed.
>
> ... Or the other way around (standardized |hvc0|, use |ttyS0| when 
> needed), which I agree.
>
That is how it is now, right?  hvc0 is the default, but that default 
seems to be overridden in rather a lot of places.

> Anyway, you shouldn't use two identical files for different providers. 
> The |ConfigureGrub| task from both 
> |bootstrapvz/providers/kvm/tasks/boot.py| and 
> |bootstrapvz/providers/virtualbox/tasks/boot.py| could be perfectly 
> moved to |bootstrapvz/common/tasks/boot.py| (and imported/added to the 
> taskset in their |__init__.py| when needed, as you are aware of).
>
> Regards,
> Tiago.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub 
> <https://github.com/andsens/bootstrap-vz/pull/320#issuecomment-218037997>
>


---
name: debian-jessie
provider:
  name: kvm
bootstrapper:
  workspace: /target
image:
  name: debian-{system.release}-{system.architecture}-{%y}{%m}{%d}
  description: Debian {system.release} {system.architecture}
system:
  release: jessie
  architecture: amd64
  bootloader: grub
  charmap: UTF-8
  locale: en_US
  timezone: UTC
volume:
  backing: raw
  partitions:
    type: gpt
    root:
      filesystem: ext4
      size: 19GiB
    swap:
      size: 512MiB
packages:
  sources:
    wikimedia:
      - deb  http://apt.wikimedia.org/wikimedia jessie-wikimedia main backports thirdparty
      - deb-src http://apt.wikimedia.org/wikimedia jessie-wikimedia main backports thirdparty
  trusted-keys:
    - /etc/apt/trusted.gpg
  install_standard: true
  install:
    - linux-image-3.19.0-2-amd64
    - linux-meta
    - coreutils
    - snmp
    - wipe
    - tzdata
    - zsh-beta
    - jfsutils
    - xfsprogs
    - screen
    - salt-minion
    - gdb
    - iperf
    - htop
    - vim
    - sysstat
    - ngrep
    - acct
    - git
    - lldpd
    - parted
    - lvm2
    - emacs
    - nslcd
    - nslcd-utils
    - libpam-ldapd
    - ldap-utils
    - libnss-ldapd
    - nss-updatedb
    - libnss-db
    - nscd
    - libpam-ldapd
    - python-ldap
    - python-pycurl
    - openssl
    - ca-certificates
    - ssl-cert
    - rsyslog
    - exim4-config
    - exim4-daemon-light
    - cloud-utils
    - euca2ools
    - openssh-server
    - curl
    - nfs-client
    - tree
    - libfile-next-perl
    - ack-grep
    - puppet-lint
    - ntp
    - debian-goodies
    - python-redis
    - atop
    - virt-what
    - pv
    - ldapvi
    - ack-grep
    - ethtool
    - debian-goodies
    - python-redis
    - python-apt
    - apt-show-versions
    - pv
    - nagios-plugins-standard
    - httpry
    - nagios-plugins
    - tree
    - nagios-plugins-basic
    - jq
    - debdeploy-minion
    - arcconf
    - megacli
    - apt-transport-https
    - tcpdump
    - molly-guard
    - tshark
    - tmux
    - strace
    - python-statsd
    - git-fat
    - nagios-nrpe-server
# Puppet installs mpt-status right away, but I'm told that installing
#  it in a raw image causes it to send spam mails.
#    - mpt-status
plugins:
  root_password:
    password: test
  cloud_init:
    username: admin
    metadata_sources: ConfigDrive
  puppet:
    assets: /etc/bootstrap-vz/puppet
  file_copy:
    files:
      -
        src: /etc/bootstrap-vz/firstscripts/firstboot.sh
        dst: /root/firstboot.sh
        permissions: "755"
        owner: root
      -
        src: /etc/bootstrap-vz/firstscripts/firstbootrc
        dst: /etc/rc.local
        permissions: "755"
        owner: root
      -
        src: /etc/bootstrap-vz/cloud.cfg
        dst: /etc/cloud/cloud.cfg
      -
        src: /etc/bootstrap-vz/access.conf
        dst: /etc/security/access.conf
      -
        src: /etc/ldap.conf
        dst: /etc/ldap.conf
      -
        src: /etc/bootstrap-vz/nss_ldap.conf
        dst: /etc/ldap/ldap.conf
      -
        src: /etc/nscd.conf
        dst: /etc/nscd.conf
      -
        src: /etc/bootstrap-vz/nslcd.conf
        dst: /etc/nslcd.conf
      -
        src: /etc/nsswitch.conf
        dst: /etc/nsswitch.conf
      -
        src: /etc/pam.d/common-auth
        dst: /etc/pam.d/common-auth
      -
        src: /etc/pam.d/sshd
        dst: /etc/pam.d/sshd
      -
        src: /etc/pam.d/common-account
        dst: /etc/pam.d/common-account
      -
        src: /etc/pam.d/common-password
        dst: /etc/pam.d/common-password
      -
        src: /etc/pam.d/common-session
        dst: /etc/pam.d/common-session
      -
        src: /etc/pam.d/common-session-noninteractive
        dst: /etc/pam.d/common-session-noninteractive
      -
        src: /etc/exim4/exim4.conf
        dst: /etc/exim4/exim4.conf
      -
        src: /etc/sudoers
        dst: /etc/sudoers
      -
        src: /etc/sudoers.d/ops
        dst: /etc/sudoers.d/ops
      -
        src: /etc/ssh/sshd_config
        dst: /etc/ssh/sshd_config
      -
        src: /etc/ssh/sshd_banner
        dst: /etc/ssh/sshd_banner
      -
        src: /etc/apt/trusted.gpg
        dst: /etc/apt/trusted.gpg
      -
        src: /etc/apt/sources.list.d/wikimedia.list
        dst: /etc/apt/sources.list.d/wikimedia.list
      -
        src: /etc/default/grub
        dst: /etc/default/grub
      -
        src: /etc/apt/preferences.d/wikimedia.pref
        dst: /etc/apt/preferences.d/wikimedia.pref
      -
        src: /etc/ssl/certs/ca-certificates.crt
        dst: /etc/ssl/certs/ca-certificates.crt
  commands:
    # Current implementation does not specify an order between image_commands and
    #  file_copy.  This section should not depend on anything existing from
    #  file_copy; hence the touch below.
    commands:
      - ['chroot', '{root}', 'passwd', '-ld', 'root']
      - ['chroot', '{root}', '/bin/sh', '-c', 'echo "#" >> /etc/sudoers']
      - ['chroot', '{root}', '/bin/sh', '-c', 'touch /etc/sudoers && SUDO_FORCE_REMOVE=yes apt-get --force-yes --assume-yes install sudo-ldap']
      - ['chroot', '{root}', 'rm', '-f', '/etc/ssh/ssh_host*key*']
      - ['chroot', '{root}', 'rm', '-f', '/etc/sudo-ldap.conf']
      - ['chroot', '{root}', 'ln', '-s', '/etc/ldap/ldap.conf', '/etc/sudo-ldap.conf']
      - ['chroot', '{root}', 'rm', '-f', '/etc/resolvconf/resolv.conf.d/original']
      - ['chroot', '{root}', 'puppet', 'agent', '--disable']

  unattended_upgrades:
    update_interval: 1
    download_interval: 1
    upgrade_interval: 1
  cloud_init:
    metadata_sources: Ec2
    username: admin

myhro:
Andrew,

> I've updated #318 with a wordier statement of my problem.  I'm attaching the full version of the
manifest that produces the issue.

Thank you. Now I could reproduce the issue. The difference was in the way we were starting the VM (e.g. using GUI vs. CLI).

> I would be happy to try and reproduce the bug with that manifest if you send it my way. I don't see it in my git clone.

Sorry. I've forgot to mention that I added it a little before replying you, but there's no need. The problem isn't in the manifest, as I mentioned above.

> That is how it is now, right?  hvc0 is the default, but that default seems to be overridden in rather a lot of places.

Now that you mentioned, looks like `ttyS0` actually isn't, but should be the default, as `hvc0` was introduced in 834ef3d for EC2 and then used by `bootstrapvz/common/tasks/grub.py` since 608de63. At least Azure and GCE are switching back to `ttyS0` and this is causing problems with KVM (which I confirmed) and possibly VirtualBox.

The solution I would like to see for this problem is an optional `console` entry for `grub` in the manifest, where `ttyS0` is the default but could be changed to `hvc0` when needed. The problem is that it would involve dropping the places where this is done separately. Are you able to do this?

If we can't do this (for time and manpower constraints), maybe we can change `hvc0` back to `ttyS0` on `bootstrapvz/common/tasks/grub.py`. What do you think, @andsens ?

andrewbogott:
On 5/9/16 10:32 PM, Tiago Ilieve wrote:
>
> Andrew,
>
>     I've updated #318
>     <https://github.com/andsens/bootstrap-vz/issues/318> with a
>     wordier statement of my problem. I'm attaching the full version of the
>     manifest that produces the issue.
>
> Thank you. Now I could reproduce the issue. The difference was in the 
> way we were starting the VM (e.g. using GUI vs. CLI).
>
great!
>
>     I would be happy to try and reproduce the bug with that manifest
>     if you send it my way. I don't see it in my git clone.
>
> Sorry. I've forgot to mention that I added it a little before replying 
> you, but there's no need. The problem isn't in the manifest, as I 
> mentioned above.
>
>     That is how it is now, right? hvc0 is the default, but that
>     default seems to be overridden in rather a lot of places.
>
> Now that you mentioned, looks like |ttyS0| actually isn't, but should 
> be the default, as |hvc0| was introduced in 834ef3d 
> <https://github.com/andsens/bootstrap-vz/commit/834ef3d58c5ba8322bd2f804f9e7f51aa62cf559> 
> for EC2 and then used by |bootstrapvz/common/tasks/grub.py| since 
> 608de63 
> <https://github.com/andsens/bootstrap-vz/commit/608de63d3ed4a365d6409007805d73d1f2ea909a>. 
> At least Azure and GCE are switching back to |ttyS0| and this is 
> causing problems with KVM (which I confirmed) and possibly VirtualBox.
>
> The solution I would like to see for this problem is an optional 
> |console| entry for |grub| in the manifest, where |ttyS0| is the 
> default but could be changed to |hvc0| when needed. The problem is 
> that it would involve dropping the places where this is done 
> separately. Are you able to do this?
>
Would that be an arg in the 'system' section?  I ask because as far as I 
can see, the only place that 'grub' is specified in  a manifest is as 
the bootloader option, which has alternative settings besides 'grub'.  
So we could add an optional grub-console arg under 'system' but it 
wouldn't always apply.


> If we can't do this (for time and manpower constraints), maybe we can 
> change |hvc0| back to |ttyS0| on |bootstrapvz/common/tasks/grub.py|.
>
That seems right to me, although I worry that there are now use cases in 
the wild that depend on it being set to hvc0 since it's been that way 
for a while :(

andsens:
> The solution I would like to see for this problem is an optional console entry for grub in the manifest, where ttyS0 is the default but could be changed to hvc0 when needed. The problem is that it would involve dropping the places where this is done separately. Are you able to do this?

> If we can't do this (for time and manpower constraints), maybe we can change hvc0 back to ttyS0 on bootstrapvz/common/tasks/grub.py. What do you think, @andsens ?

I disagree with the solution. The user shouldn't have to know which console setting to choose for grub, it should be automatic. There should be a `ConfigureGrubConsoleTTY0` task, which is added per default through the `boot` group and a `ConfigureGrubConsoleHVC0` task which is added through the ec2 provider (and which discards the TTY0 task).

myhro:
Anders,

> I disagree with the solution. The user shouldn't have to know which console setting to choose for grub, it should be automatic.

I was thinking about the use case of GUI clients for KVM/VirtualBox, where would make sense to see/use the GRUB menu. Of course this is not for the main proportion of bootstrap-vz users, but at least someone who needs it would have an option (and not being actually forced to set the console in the manifest). That's why I suggested an `optional` entry for this, which would be automatic for anyone who doesn't care about it.

andrewbogott:
On 5/10/16 10:43 AM, Tiago Ilieve wrote:
>
> Anders,
>
>     I disagree with the solution. The user shouldn't have to know
>     which console setting to choose for grub, it should be automatic.
>
> I was thinking about the use case of GUI clients for KVM/VirtualBox, 
> where would make sense to see/use the GRUB menu. Of course this is not 
> for the main proportion of bootstrap-vz users, but at least someone 
> who needs it would have an option (and not being actually forced to 
> set the console in the manifest). That's why I suggested an |optional| 
> entry for this, which would be automatic for anyone who doesn't care 
> about it.
>

We may be a bit too far into the weeds on this.  I don't object to 
making some kind of pluggable fix, but we can get working with a /much/ 
simpler patch.  The ec2 provider already overrides to console=hvc0 in 
ConfigurePVGrub().  So simply changing the default to ttyS0 should cover 
all the cases that I know about.  New tiny pull request is here: 
https://github.com/andsens/bootstrap-vz/pull/321

myhro:
I'm closing this as #321 was merged.

PR #321, opened 2016-05-10
Set a default of GRUB_CMDLINE_LINUX_DEFAULT="console=ttyS0"

andrewbogott:
This needed changing to hvc0 for ec2, and was accidentally changed
for all other providers.  Since the ec2 provider overrides to hvc0
anyway, we can just change this default to the setting that works
for everyone else.

Fixes: #318


myhro:
Andrew,

Indeed. This will also not break [Azure][1] and [GCE][2] implementations, as they substitute the entire `GRUB_CMDLINE_LINUX*` line (i.e., they don't expect an `hvc0` entry to be overrided).

Many thanks for taking care of this!

Merging...

[1]: https://github.com/andsens/bootstrap-vz/blob/bc1d546/bootstrapvz/providers/azure/tasks/boot.py#L34
[2]: https://github.com/andsens/bootstrap-vz/blob/bc1d546/bootstrapvz/providers/gce/tasks/boot.py#L16

Issue #91, opened 2014-05-10
Copying EC2 images over regions.


dbalan:
Since AWS provides an abstraction to copy the images generated over mutliple regions, doesn't it make sense to expose it in the manifest?

I have a written a [simple plugin in this commit](https://github.com/plivo/bootstrap-vz/commit/3c1f4f4cd8bf473bfee13d30e7c4de9a1d90788a), but think it would be better to go into the `image` section in the `manifest`. What are your thoughts?

PR #327, opened 2016-06-02
ec2_publish plugin

### LINKED ISSUE #271: EC2 ami build should return AMI id
It would be very helpful if bootstrap-vz can return AMI id it just generated. 
Will help with automated builds. (Packer does it, but it is not well Debian aware ...)

thepwagner:
Fixes #271, #91

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,10 +1,16 @@
 Changelog
 =========
 
+
+2016-06-02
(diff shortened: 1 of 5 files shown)


andsens:
Merged manually. Thank you, I think @JamesBromberger will appreciate this plugin a lot, he maintains the official images :-)

PR #332, opened 2016-08-12
docker: Fix quoting that breaks some versions of Docker

KellerFuchs:
The extra quoting breaks on some versions of Docker, such as the one available on Travis.

See <https://travis-ci.org/hashbang/docker-gitolite/builds/151947017#L527> for a build log.

### DIFF
--- a/bootstrapvz/providers/docker/tasks/image.py
+++ b/bootstrapvz/providers/docker/tasks/image.py
@@ -50,15 +50,9 @@ class PopulateLabels(Task):
             for label, value in info.manifest.provider['labels'].items():
                 labels[label] = value.format(**info.manifest_vars)
 
-        # pipes.quote converts newlines into \n rather than just prefixing
-        # it with a backslash, so we need to escape manually
-        def escape(value):
-            value = value.replace('"', '\\"')
-            value = value.replace('\n', '\\\n')
-            value = '"' + value + '"'
-            return value
+        from pipes import quote
         for label, value in labels.items():
-            info._docker['dockerfile'].append('LABEL {}={}'.format(label, escape(value)))
+            info._docker['dockerfile'].append('LABEL {}={}'.format(label, quote(value)))
 
 
 class AppendManifestDockerfile(Task):


andsens:
According to [the documentation](https://docs.docker.com/engine/reference/builder/#/label), we should
> use quotes and backslashes as you would in command-line parsing.

i.e. this should work (and it has, something in the docker internals changed). The escape function itself is however very sub-par, I have no clue why I didn't use [`pipes.quote()`](https://docs.python.org/2/library/pipes.html#pipes.quote) instead. Could you do that and see if it makes any difference?

Your change would break labels that have spaces in them.

KellerFuchs:
>  I have no clue why I didn't use pipes.quote() instead.

You left a comment there:
> \# pipes.quote converts newlines into \n rather than just prefixing
> \# it with a backslash, so we need to escape manually

However, I just replaced this with `pipes.quote` (it is deprecated, but `shlex.quote` is Python 3.3+) and it seems to work perfectly:
<https://travis-ci.org/hashbang/docker-gitolite/builds/152042273>

andsens:
Hah, what good are comments if I don't read them myself? :-D

Looks good. Did you try a label with newlines in it? After all, that's what my comment suggests would break stuff.

KellerFuchs:
@andsens Yes, I forgot to mention it be setting this label worked:

KellerFuchs:
P.P.S: According to [Docker's documentation](https://docs.docker.com/engine/userguide/labels-custom-metadata/#/add-labels-to-images), it doesn't seem possible to express `LABEL`s with linebreaks anyway, so that might be a limitation of Docker  :V

andsens:
Welp, you're right! OK then, merging.

Thank you @KellerFuchs, nice job :-)

KellerFuchs:
Well, thank you @andsens.

So far, I've had an overwhelmingly positive experience with bootstrap-vz, and now I have pretty much feature-parity (in terms of CI/CD automation and pushing to Dockerhub) with Dockerfiles  :-)

andsens:
That's wonderful to hear. Positive feedback never fails to make my day, so thank you :-)

KellerFuchs:
@andsens Well, you might hear more of it in the near future, since @lrvick and I might move the @hashbang container stuff to `bootstrap-vz`.
That or get more PRs.  😁

PR #333, opened 2016-08-13
apt: Validate packages.trusted-keys

nbraud:
This makes wrong (inexistent or malformed) trusted-keys
keyring fail the image build at validation-time, rather
than in the package installation phase.

### DIFF
--- a/bootstrapvz/common/phases.py
+++ b/bootstrapvz/common/phases.py
@@ -1,5 +1,6 @@
 from bootstrapvz.base.phase import Phase
 
+validation = Phase('Validation', 'Validating data, files, etc.')
 preparation = Phase('Preparation', 'Initializing connections, fetching data etc.')
 volume_creation = Phase('Volume creation', 'Creating the volume to bootstrap onto')
 volume_preparation = Phase('Volume preparation', 'Formatting the bootstrap volume')
@@ -13,7 +14,8 @@ volume_unmounting = Phase('Volume unmounting', 'Unmounting the bootstrap volume'
 image_registration = Phase('Image registration', 'Uploading/Registering with the provider')
 cleaning = Phase('Cleaning', 'Removing temporary files')
 
-order = [preparation,
+order = [validation,
+         preparation,
          volume_creation,
          volume_preparation,
          volume_mounting,
--- a/bootstrapvz/common/task_groups.py
+++ b/bootstrapvz/common/task_groups.py
@@ -121,6 +121,7 @@ def get_apt_group(manifest):
     if 'sources' in manifest.packages:
         group.append(apt.AddManifestSources)
     if 'trusted-keys' in manifest.packages:
+        group.append(apt.ValidateTrustedKeys)
         group.append(apt.InstallTrustedKeys)
     if 'preferences' in manifest.packages:
         group.append(apt.AddManifestPreferences)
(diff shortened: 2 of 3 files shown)


andsens:
Nice stuff! As much as possible should be checked up-front :-)

However, I believe it would be better to put this into a task that runs during the preparation phase. `validate_manifest` is mostly meant for checking internal consistency of the manifest (or that is the intention, I have not really documented that anywhere yet).

nbraud:
> `validate_manifest` is mostly meant for checking internal consistency of the manifest (or that is the intention, I have not really documented that anywhere yet).

Ah, ok.  Is there any specific reason for this?

andsens:
> Ah, ok. Is there any specific reason for this?

Well, kind of. The function was introduced as a stop-gap measure for validation that wasn't possible or too complicated with json schemas. The reason `validate_manifest` only receives the raw data is exactly because we don't want to create a valid manifest model before we are sure of its internal consistency.

> As far as I understand, tasks form a dependency graph (where any plugin/provider/... can introduce new dependencies), and validation doesn't have dependency/ordering constraints.

That's fine though, then they'll just run at a random point during whichever phase they're in. Task discovery happens way before any ordering, so you don't have to worry about them not being included.

nbraud:
@andsens OK, then I will redo the whole thing with a validation phase instead. (same for #335)

andsens:
Cool! I appreciate the effort. Once we got that phase rolling I'm sure I can find some more tasks that can be moved there (e.g. [CheckAptProxy](https://github.com/andsens/bootstrap-vz/blob/b9d63b701df3e7757a6f3c3336ce2cc71af6b4b2/bootstrapvz/plugins/apt_proxy/tasks.py#L8) and [CheckBoxPath](https://github.com/andsens/bootstrap-vz/blob/b9d63b701df3e7757a6f3c3336ce2cc71af6b4b2/bootstrapvz/plugins/vagrant/tasks.py#L10)).

nbraud:
@andsens Updated.  The “common: Add a validation phase” commit is the same as for #335, so they can be merged in whichever order.

Something I'm not sure about is whether calling `info.manifest.validation_error` in the `validate` phase is legit.

nbraud:
@andsens I filed #339 to move the relevant tasks in the `validation` phase.

andsens:
Nice :-)

> Something I'm not sure about is whether calling info.manifest.validation_error in the validate phase is legit.

I never thought of using it in a task, but it makes perfect sens to pinpoint the error in a manifest like this. The same should probably be done for the other validation errors :-)

Merging...

justinsb:
This seems to have broken bootstrap-vz for me.  I am trying to validate a GPG key I generated using these build instructions

gpg: WARNING: unsafe permissions on homedir `/tmp/tmpESNoIA'
gpg: [don't know]: invalid packet (ctb=2d)
gpg: keydb_search_first failed: invalid packet

nbraud:
@justinsb Yes, this seems to be a key format issue:
> file /tmp/kopeio.gpg.key
> /tmp/kopeio.gpg.key: PGP public key block Public-Key (old)

I was able to import, re-export, then validate it.
I will check how to fix this.

justinsb:
I am likely exporting in an "old" way - sorry & thanks for looking at it!  What is the correct way to export the key?

nbraud:
@justinsb In my case, `gpg2 --export -a $keyid` did the right thing.
As a FYI, you probably want to use the flag `--export-options export-clean,export-minimal`.

In any case, the validation is wrong here: this is something which is accepted by `apt`, bootstrap-vz shouldn't reject it.

PR #331, opened 2016-08-07
Misc. fixes & improvements

nbraud:
- ~~admin_user: Allow designating pubkeys by relative path~~
- file_copy: Allow copying directories
- apt: Don't add a security mirror for obsolete releases  
  This made the build of a squeeze image (for instance) fail.

PR #335, opened 2016-08-13
admin_user & file_copy: Make paths relative to the manifest

nbraud:
This is a followup on #331, since this specific patch was dropped.

@andsens I agree that paths relative to the manifest would be best.
However, while implementing this I ran into trouble: there is no way for `validate_manifest` to access `manifest.path`.
Moreover, the `file_copy` plugin also allows relative paths, and has the same issue (pathes end up being relative to the working directory).
I suggest that we pass the entire `manifest` object to `validate_manifest`, rather than `manifest.data`, `manifest.schema_validator` and `manifest.validation_error`.

### DIFF
--- a/bootstrapvz/common/tools.py
+++ b/bootstrapvz/common/tools.py
@@ -133,3 +133,8 @@ def copy_tree(from_path, to_path):
         for path in files:
             copy(os.path.join(abs_prefix, path),
                  os.path.join(to_path, prefix, path))
+
+
+def rel_path(base, path):
+    import os.path
+    return os.path.join(os.path.dirname(base), path)
--- a/bootstrapvz/plugins/admin_user/__init__.py
+++ b/bootstrapvz/plugins/admin_user/__init__.py
@@ -1,13 +1,8 @@
-
-
 def validate_manifest(data, validator, error):
     import os.path
+
     schema_path = os.path.normpath(os.path.join(os.path.dirname(__file__), 'manifest-schema.yml'))
     validator(data, schema_path)
-    pubkey = data['plugins']['admin_user'].get('pubkey', None)
-    if pubkey is not None and not os.path.exists(pubkey):
-        msg = 'Could not find public key at %s' % pubkey
-        error(msg, ['plugins', 'admin_user', 'pubkey'])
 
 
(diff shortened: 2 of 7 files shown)


nbraud:
@andsens This is my proposed solution:
- pass the whole manifest to `validate_manifest()`;
- make paths relatives to the manifest (in `admin_user` and `file_copy`), and validate them properly.

Note that the `file_copy` change is breaking: the documentation previously asserted that paths were relative to the “current directory”.

andsens:
Nice work :-)
I think however a better idea would be to add those validations as tasks, there you have full access to everything that is needed. It might be a good idea to actually have a new "validation" phase before the "preparation" phase, which checks all the external dependencies of the manifest.

nbraud:
@andsens Updated, and I sneaked in a bugfix

bootstrapvz/common/tools.py
```diff
@@ -133,3 +133,12 @@ def copy_tree(from_path, to_path):
         for path in files:
             copy(os.path.join(abs_prefix, path),
                  os.path.join(to_path, prefix, path))
+
+
+def rel_path(base, path):
```
andsens: There's actually no need for this one, check this out:
```
$ python
>>> import os
>>> os.path.join('/some/absolute/path', '/another/absolute/path')
'/another/absolute/path'
```
:-)

andsens:
I added a small comment, otherwise this looks great!

nbraud:
Thanks

andsens:
> However, while implementing this I ran into trouble: there is no way for validate_manifest to access manifest.path.

Oh crap. There's a reason for this and I totally forgot about it during this entire PR. You can do remote bootstrapping where the manifest is actually located on another host! I actually have no clue how paths would have worked previously at all.

We definitely need this PR, it's awesome work. It's late though and I need some time and a clear mind to think through the consequences before merging, cool?

nbraud:
@andsens No problem, we aren't on a deadline anyhow.

andsens:
OK. Thought about it. Right now, this would result in an error when using remote bootstrapping. I'll fix it when I get to it. Merging :-)

PR #339, opened 2016-09-11
Move checking tasks to the validation phase

nbraud:
This moves existing tasks that only perform extended validation (file exists, server is reachable, ...) to the new `validation` phase.


andsens:
Yes! Magnificent :-)

Merging.

PR #341, opened 2016-09-12
Use the rel_path utility function

nbraud:
This piggy-backs on #335 and refactors the existing path-handling code using the new helper function.

Merge #335 first.


nbraud:
The code got easier to read and we get more uniformity: `os.path.normpath` is called systematically, for instance.

nbraud:
@andsens Looks like 90da634226f11186685c22edb90a77f7f111da82 ran in a non-deterministic test failure.

PR #342, opened 2016-09-20
Fix cloud-init config

JamesBromberger:
Hi Anders,
This pull fixes cloud-init from overwriting the apt sources.list file, as well as enabling cloud-init to manage (update) the /etc/hosts file with the local hostname. I'm generating 8.6 AMIs with this, and they look pretty good.


andsens:
Cool. LGTM 👍

PR #343, opened 2016-09-21
Plugin: ec2_launch ssh key, instance type, etc. +

vorlock:
* added ssh_key parameter to the manifest to connect with (have to be
  present in AWS)
* added readme
* changed instance type to something what can run outside VPC (m3.medium)


andsens:
LGTM 👍 
The test failure is nothing to worry about, it's a timing issue with the test itself.

PR #344, opened 2016-09-22
Finishing touches for Stretch

vorlock:
This set of patches allows to create Stretch AMI and should also be a good background for Stretch for other providers

### DIFF
--- a/bootstrapvz/common/tasks/ssh.py
+++ b/bootstrapvz/common/tasks/ssh.py
@@ -101,6 +101,7 @@ class ShredHostkeys(Task):
     def run(cls, info):
         ssh_hostkeys = ['ssh_host_dsa_key',
                         'ssh_host_rsa_key']
+
         from bootstrapvz.common.releases import wheezy
         if info.manifest.release >= wheezy:
             ssh_hostkeys.append('ssh_host_ecdsa_key')
@@ -109,4 +110,5 @@ class ShredHostkeys(Task):
         public = [path + '.pub' for path in private]
 
         from ..tools import log_check_call
-        log_check_call(['shred', '--remove'] + private + public)
+        log_check_call(['shred', '--remove'] + [key for key in private + public
+                                                if os.path.isfile(key)])
--- a/bootstrapvz/providers/ec2/__init__.py
+++ b/bootstrapvz/providers/ec2/__init__.py
@@ -49,7 +49,7 @@ def validate_manifest(data, validator, error):
 
 
 def resolve_tasks(taskset, manifest):
-    from bootstrapvz.common.releases import wheezy, jessie
+    from bootstrapvz.common.releases import wheezy, jessie, stable
 
     taskset.update(task_groups.get_standard_groups(manifest))
     taskset.update(task_groups.ssh_group)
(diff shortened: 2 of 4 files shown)


bootstrapvz/common/tasks/ssh.py
```diff
@@ -109,4 +110,8 @@ def run(cls, info):
         public = [path + '.pub' for path in private]
 
         from ..tools import log_check_call
-        log_check_call(['shred', '--remove'] + private + public)
+        for ssh_host_key in private + public:
+            if os.path.isfile(ssh_host_key):
```
andsens: Why the change? Do some of the keys not exist? In that case they should be removed from the list, we should always strive to know what *exactly* is on the image, so that we can fail if something is missing unexpectedly.

Also, `ssh_host_keys` should probably be distinguished form the other variable with something more than just an underscore. Alternatively, simply negate the `isfile()` and remove the key from the list.

bootstrapvz/providers/ec2/__init__.py
```diff
@@ -80,12 +80,13 @@ def resolve_tasks(taskset, manifest):
         taskset.add(tasks.network.EnableDHCPCDDNS)
 
     if manifest.release >= jessie:
-        taskset.add(apt.AddBackports)
         taskset.add(tasks.tuning.SetCloudInitMountOptions)
         taskset.add(tasks.packages.AddWorkaroundGrowpart)
         taskset.add(initd.AdjustGrowpartWorkaround)
         if manifest.system['bootloader'] == 'grub':
             taskset.add(grub.EnableSystemd)
+        if manifest.release == stable:
```
andsens: This change needs a little explanation.

vorlock: This possibly could be `if manifest.release <= stable:` as backports are not available for testing/sid.

bootstrapvz/common/tasks/ssh.py
vorlock: Yeah DSA keys are not in Stretch but it is not set in stone that somebody will not use them and to be honest we should remove all keys which are generated during bootstrapping. So I'd say in this case removing all of them silently is not a big issue.

You're right it could be clearer is `host_key` ok with you?

andsens: Why not just:
```py
log_check_call(['shred', '--remove'] + [key for key in private + public if os.path.isfile(key))
```

bootstrapvz/providers/ec2/__init__.py
andsens: Ah, ok. The reason backports was added is actually in order to get `cloud-init` working, I think. So you could actually just make it `manifest.release == jessie`, it was added in #256, where there are a slew of other options added, so the justification for adding backports in general seems a little thin.

bootstrapvz/common/tasks/ssh.py
vorlock: Yes, your piece of code is more pythonic :-) I just changed it and rebased last commit.

bootstrapvz/providers/ec2/__init__.py
vorlock: I think I prefer to stick to stable for simple reason in the future we may need backports for the future stable and this would require changes in the code again, with what I'm suggesting we're always having it available if needed.

andsens: Before merging, could you change this to `<= stable`? Otherwise backports will disappear on jessie images once stretch is stable.

vorlock: Done as requested rebased on top of master and pushed

andsens:
Perfect :-)

PR #346, opened 2016-09-25
Added tagging for the created images +

vorlock:
- fully implemented for EC2 provider only, other providers will require
  validation and applying to the instances added
- brand new Stretch manifest with top level AMI tags
- some minor Jessie manifest editorials
- some minor yaml schema editorials

### DIFF
--- a/bootstrapvz/base/manifest-schema.yml
+++ b/bootstrapvz/base/manifest-schema.yml
@@ -47,10 +47,10 @@ properties:
         enum: [i386]
       bootloader:
         enum:
-        - pvgrub
-        - grub
-        - extlinux
-        - none
+          - pvgrub
+          - grub
+          - extlinux
+          - none
       charmap: {type: string}
       hostname:
         type: string
@@ -59,12 +59,12 @@ properties:
       release: {type: string}
       timezone: {type: string}
     required:
-    - release
-    - architecture
-    - bootloader
-    - timezone
-    - locale
-    - charmap
+      - release
+      - architecture
+      - bootloader
+      - timezone
+      - locale
+      - charmap
     additionalProperties: false
   packages:
     type: object
@@ -77,8 +77,8 @@ properties:
         type: array
         items:
           anyOf:
-          - pattern: ^[^/]+(/[^/]+)?$
-          - $ref: '#/definitions/absolute_path'
+            - pattern: ^[^/]+(/[^/]+)?$
+            - $ref: '#/definitions/absolute_path'
         minItems: 1
       install_standard: {type: boolean}
       mirror:
@@ -137,8 +137,8 @@ properties:
       partitions:
         type: object
         oneOf:
-        - $ref: '#/definitions/no_partitions'
-        - $ref: '#/definitions/partition_table'
+          - $ref: '#/definitions/no_partitions'
+          - $ref: '#/definitions/partition_table'
     required: [partitions]
     additionalProperties: false
 definitions:
--- a/bootstrapvz/providers/ec2/__init__.py
+++ b/bootstrapvz/providers/ec2/__init__.py
@@ -49,6 +49,9 @@ def validate_manifest(data, validator, error):
 
 
 def resolve_tasks(taskset, manifest):
+    """
+    Function setting up tasks to run for this provider
+    """
     from bootstrapvz.common.releases import wheezy, jessie, stable
 
     taskset.update(task_groups.get_standard_groups(manifest))
--- a/bootstrapvz/providers/ec2/tasks/ami.py
+++ b/bootstrapvz/providers/ec2/tasks/ami.py
@@ -125,3 +125,9 @@ class RegisterAMI(Task):
             registration_params['sriov_net_support'] = 'simple'
 
         info._ec2['image'] = info._ec2['connection'].register_image(**registration_params)
+
+        # Setting up tags on the AMI
+        if 'tags' in info.manifest.data:
+            raw_tags = info.manifest.data['tags']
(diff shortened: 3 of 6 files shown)


bootstrapvz/providers/ec2/tasks/ami.py
```diff
@@ -125,3 +125,10 @@ def run(cls, info):
             registration_params['sriov_net_support'] = 'simple'
 
         info._ec2['image'] = info._ec2['connection'].register_image(**registration_params)
+
+        # Setting up tags on the AMI
+        if info.manifest.tags:
```
andsens: Just for completeness sake you should probably check for `is not None` here, this way it's possible to know exactly which type the variable is in the negative case. Alternatively, you could make it an empty `dict` during init and check with `len() > 0` instead.  

Also: Why the `else: pass`?

manifests/official/ec2/ebs-stretch-amd64-hvm.yml
```diff
@@ -0,0 +1,68 @@
+---
+name: debian-{system.release}-{system.architecture}-{provider.virtualization}-{%Y}-{%m}-{%d}-{%H}{%M}-ebs
+tags: {
```
andsens: No need for the braces.

bootstrapvz/base/manifest-schema.yml
```diff
@@ -47,10 +48,10 @@ properties:
...
       bootloader:
         enum:
-        - pvgrub
-        - grub
-        - extlinux
-        - none
+          - pvgrub
+          - grub
+          - extlinux
+          - none
```
andsens: 👍  for this change, both versions are valid, but I really don't like the unindented version either, it's ambiguous.

bootstrapvz/base/manifest-schema.yml
```diff
@@ -6,6 +6,7 @@ required: [name, provider, bootstrapper, system, volume]
 properties:
   name:
     type: string
+  tags: {type: object}
```
andsens: You should put this in the provider section. We should rarely need to modify the root properties of a manifest. Tags are something provider specific, even though the same concept pops up on most of them, their usage and implementation can be very different.

bootstrapvz/providers/ec2/tasks/ami.py
vorlock: when if was failing something was causing whole process to crash, TBH I don't remember now what and why it was like this, it's being rewritten anyway

manifests/official/ec2/ebs-stretch-amd64-hvm.yml
vorlock: ack

bootstrapvz/base/manifest-schema.yml
vorlock: I agree and migrated whole tagging thing to EC2 provider

vorlock:
@andsens I just pushed changed commit incorporating your suggestions

bootstrapvz/providers/ec2/__init__.py
```diff
@@ -47,8 +48,15 @@ def validate_manifest(data, validator, error):
     if enhanced_networking == 'simple' and virtualization != 'hvm':
             error('Enhanced networking only works with HVM virtualization', ['provider', 'virtualization'])
 
+    # Checking if 'tags' is a dict
```
andsens: There is no need for this check. The manifest does that for you.

bootstrapvz/providers/ec2/tasks/ami.py
```diff
@@ -125,3 +125,12 @@ def run(cls, info):
             registration_params['sriov_net_support'] = 'simple'
 
         info._ec2['image'] = info._ec2['connection'].register_image(**registration_params)
+
+        # Setting up tags on the AMI
+        if 'tags' in info.manifest.data and len(info.manifest.data['tags']) > 0:
+            def apply_format(v):
```
andsens: This lambda seems unnecessary, all you are doing is currying one argument, better to inline it to improve readability. Great idea to use the manifest vars btw. it's very applicable in this case.

bootstrapvz/providers/ec2/tasks/ami.py
```diff
@@ -125,3 +125,12 @@ def run(cls, info):
             registration_params['sriov_net_support'] = 'simple'
 
         info._ec2['image'] = info._ec2['connection'].register_image(**registration_params)
+
+        # Setting up tags on the AMI
+        if 'tags' in info.manifest.data and len(info.manifest.data['tags']) > 0:
```
andsens: You can enforce the `len(info.manifest.data['tags']) > 0` by setting a min length in the manifest, this way we move as much validation as possible out of the code.

bootstrapvz/providers/ec2/tasks/ami.py
vorlock: TBH I'm not sure what do you mean by the above. I can get rid of lambda but not sure if it will improve readability

andsens: No worries :-)
What I mean is that you added an entire function (`apply_format`) so you could write:
`tags = {k: apply_format(v) for k, v in raw_tags.items()}`  
instead of  
`tags = {k: v.format(**info.manifest_vars) for k, v in raw_tags.items()}`

It doesn't save you any lines and you add a layer of indirection by doing this.

vorlock:
@andsens how it looks now

andsens:
Looks awesome! Merging :-)

Issue #347, opened 2016-09-28
Exception: The path git/bootstrap-vz/bootstrap-vz/plugins/debconf/manifest-schema.yml does not point to a file


NeatNerdPrime:
* Encountered error as described in title when running a bootstrap for jessie
* discovered the cause of the error, TL;DR: debconf/__init__.py validator points to wrong file.
* will make a PR soon after reporting this bug

PR #348, opened 2016-09-28
#347 - Fix for debconf validator pointing to wrong file.

NeatNerdPrime:
* reported in https://github.com/andsens/bootstrap-vz/issues/347

  flake8: commands succeeded
  congratulations :)


andsens:
Great stuff. But actually it's the other way around. The file should be named `manifest-schema.yml` like all other manifest schemas. I made the fix myself.
Thank you for reporting it :-)

PR #350, opened 2016-10-28
Disabling PIE during ixgbevf compilation for AWS +

vorlock:
enhanced network instances, it's being activated as default in GCC6 for
Stretch upwards.


andsens:
Could you amend the commit message to explain properly

* why this change is needed,
* what PIE does and...
* what `-fno-stack-protector` means?

vorlock:
I just changed commit message and removed stack-protector change as I think I was over zealous removing it entirely after all.

andsens:
Perfect! Merging....

Issue #244, opened 2015-08-06
Partitionning: manage any mount point & mount options


josqu4red:
For instance: it is rather common to put /tmp on a dedicated partition and with noexec/nodev/nosuid mount options.
Not being limited to root/boot/swap partitions would be nice; definitions could be like:
```yaml
partitions:
  /tmp:
    filesystem: ext4
    size: 1GiB
    mount_options: defaults,nodev,noexec,nosuid
```
or with a ```mount_point``` key to keep a sane name.

Thanks in advance

andsens:
Yeah. I would like that too :-)
Right now the architecture does not allow for that. I would also like to be able to specify multiple volumes, so that you can expand diskspace individually. I'll have to come up with a model that allows for that flexibility.

PR #355, opened 2016-12-10
Add mountopts and enable definition of additional partitions

n3ph:
So far additional partitions are just working with GPT as I wasn't able get the size for the extended partition which needs to be created for all other logical ones... 

As i am not using MSDOS partitionmaps this isn't a big issue - at least for me..

### DIFF
--- a/bootstrapvz/base/manifest-schema.yml
+++ b/bootstrapvz/base/manifest-schema.yml
@@ -153,7 +153,7 @@ definitions:
     properties:
       root: {$ref: '#/definitions/partition'}
       type: {enum: [none]}
-    required: [root]
+    required: [type, root]
     additionalProperties: false
   partition:
     type: object
@@ -178,7 +178,7 @@ definitions:
           size: {$ref: '#/definitions/bytes'}
         required: [size]
       type: {enum: [msdos, gpt]}
-    required: [root]
+    required: [type, root]
     additionalProperties: false
   path:
     type: string
--- a/bootstrapvz/base/manifest-schema.yml
+++ b/bootstrapvz/base/manifest-schema.yml
@@ -164,6 +164,10 @@ definitions:
         items: {type: string}
         minItems: 1
         type: array
+      mountopts:
+        items: {type: string}
+        minItems: 1
+        type: array
       size: {$ref: '#/definitions/bytes'}
     required: [size, filesystem]
     additionalProperties: false
--- a/bootstrapvz/base/manifest-schema.yml
+++ b/bootstrapvz/base/manifest-schema.yml
@@ -173,6 +173,8 @@ definitions:
     additionalProperties: false
   partition_table:
     type: object
+    patternProperties:
+      ^(?!type|boot|root|swap).*: {$ref: '#/definitions/partition'}
     properties:
       boot: {$ref: '#/definitions/partition'}
       root: {$ref: '#/definitions/partition'}
--- a/bootstrapvz/common/tasks/filesystem.py
+++ b/bootstrapvz/common/tasks/filesystem.py
@@ -99,6 +99,33 @@ class MountBoot(Task):
         p_map.root.add_mount(p_map.boot, 'boot')
 
 
+class MountAdditional(Task):
+    description = 'Mounting additional partitions'
+    phase = phases.volume_mounting
+    predecessors = [MountRoot]
+
+    @classmethod
+    def run(cls, info):
+        import os
+        from bootstrapvz.base.fs.partitions.unformatted import UnformattedPartition
+        p_map = info.volume.partition_map
+        partitions = []
+        #  we need to get rid of UnformattedPartition and sort the list of additional partitons in order to mount them correctly
+        for partition in info.volume.partition_map.partitions:
+            if isinstance(partition, UnformattedPartition):
+                continue
+            if partition.name not in ["boot", "swap", "root"]:
+                partitions.append(partition.name)
+        for partition_name in sorted(partitions, key=len, reverse=False):
+            print partition_name
+            partition = getattr(p_map, partition_name)
+            os.makedirs(os.path.join(info.root, partition.name))
+            if partition.mountopts is None:
+                p_map.root.add_mount(getattr(p_map, partition.name), partition.name)
+            else:
+                p_map.root.add_mount(getattr(p_map, partition.name), partition.name, ['--options'] + partition.mountopts)
+
+
 class MountSpecials(Task):
     description = 'Mounting special block devices'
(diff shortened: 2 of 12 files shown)


n3ph:
Ah forgot to mention that this fixes #244

n3ph:
It seems the sphinx-build task is somehow broken.. :-/

andsens:
dayum, nice PR! And the changes are really idiomatic the the bootstrap-vz style, well done :-)

I don't have the time right now to review this though, so I apologise for the inevitable delay in merging this. I should have time during the upcoming weekend.

Also: The partition models are starting to get really fat constructors, I'm not sure how we can avoid this. All the arguments seem relevant, ideas? Additional the work done in `MountAdditional` is good, but I hate that it has to be so awkward. I should add a way to iterate through the "normal" partitions.

nbraud:
@andsens I will try reviewing this weekend or early next week, assuming I have Internet access where I'm holing up for Yule.

bootstrapvz/base/manifest-schema.yml
```diff
@@ -153,7 +153,7 @@ definitions:
     properties:
       root: {$ref: '#/definitions/partition'}
       type: {enum: [none]}
-    required: [root]
+    required: [type, root]
```
andsens: Nice catch, we definitely require `type`

bootstrapvz/base/manifest-schema.yml
```diff
@@ -164,11 +164,17 @@ definitions:
...
+        items: {type: string}
+        minItems: 1
+        type: array
       size: {$ref: '#/definitions/bytes'}
     required: [size, filesystem]
     additionalProperties: false
   partition_table:
     type: object
+    patternProperties:
+      ^(?!type|boot|root|swap).*: {$ref: '#/definitions/partition'}
```
andsens: Might as well just remove `boot` & `root` below and include them here.

bootstrapvz/common/tasks/filesystem.py
```diff
@@ -99,6 +99,32 @@ def run(cls, info):
...
+    predecessors = [MountRoot]
+
+    @classmethod
+    def run(cls, info):
+        import os
+        from bootstrapvz.base.fs.partitions.unformatted import UnformattedPartition
+        p_map = info.volume.partition_map
+        partitions = []
+        #  we need to get rid of UnformattedPartition and sort the list of additional partitons in order to mount them correctly
+        for partition in info.volume.partition_map.partitions:
```
andsens: You can replace this with some nice functional code, making the comment almost obsolete:
```py
is_additional = lambda partition: (
    not isinstance(partition, UnformattedPartition)
    and partition.name not in ["boot", "swap", "root"]
)
name_length = lambda partition: len(partition.name)
for partition in sorted(filter(is_additional, partitions), key=name_length):
    # do stuff
```


for partition_name in sorted(partitions, key=len, reverse=False):

bootstrapvz/common/tasks/filesystem.py
```diff
@@ -183,11 +210,22 @@ def run(cls, info):
                                  'dump': '1',
                                  'pass_num': '0',
                                  })
-
+        for partition in info.volume.partition_map.partitions:
```
andsens: Same as above, just make it a lambda and use a filter.

manifests/README.rst
```diff
@@ -302,6 +302,15 @@ boot, root and swap.
    -  ``{fs}``: The filesystem of the partition.
    -  ``{device_path}``: The device path of the partition.
    -  ``{size}``: The size of the partition.
+   -  ``mount_opts``: Options to mount the partition with. This optional
```
andsens: You're missing the docs for the additional mounts.

manifests/examples/kvm/stretch-virtio-partitions.yml
```diff
@@ -0,0 +1,46 @@
...
+        - nodev
+        - noexec
+        - nosuid
+        - journal_ioprio=3
+      filesystem: ext4
+      size: 1GiB
+    var:
+      filesystem: ext4
+      size: 1GiB
+    var/tmp:
```
andsens: I love the simplicity of how to specify the mountpaths, great idea just using the keys.
Do you think it would be possible to include the slash at the beginning of the path? It would make it a lot easier for people to understand what is going on at a quick glance.  
If you see `var` in that list you might think "hm, ok, I guess there are some special values", if however you see `/var` you just think "aha, paths, ok cool, I know how to fiddle with that..."

In the same vein we should probably rename `root` to `/` and `boot` to `/boot`, but I think this is plenty for one PR :-)

bootstrapvz/base/manifest-schema.yml
n3ph: definitely - could not tell why I wasn't doing this the first time..

manifests/examples/kvm/stretch-virtio-partitions.yml
n3ph: My intension in first place was to not break the syntax anywhere.. For sure - this change would be wonderful. This would generally enhance the common comprehension of the configuration.

bootstrapvz/common/tasks/filesystem.py
n3ph: Yes - I could not have done it better...

manifests/README.rst
n3ph: Docs are not my friends.. Sry :-/

bootstrapvz/base/manifest-schema.yml
n3ph: yep

manifests/README.rst
n3ph: done

n3ph:
uhm - what does travis-ci complain about?

andsens:
> uhm - what does travis-ci complain about?

My changes :-(  
https://travis-ci.org/andsens/bootstrap-vz/builds/197093466#L172-L179  
Also, [you are ending some expressions in a semicolon](https://github.com/andsens/bootstrap-vz/pull/355/commits/ff4918f1e08e04222b738c7959822d0a1779954a#diff-90c99d2816280df32c9c3ba9307ddc97R112).

manifests/examples/kvm/stretch-virtio-partitions.yml
n3ph: .. As far as I see it should not be so hard. I'll try..

n3ph:
This should be it - so far.

andsens:
Sorry for the delay, finally got to running some tests. Seems there are no problems :-)
Merging...

PR #357, opened 2017-01-10
Fixes to create stretch images. Add stretch manifest for GCE.

zmarano:
There are still some outstanding issues with the disk expansion scripts. I will look into those separately. For now I just disabled them.

### DIFF
--- a/bootstrapvz/providers/gce/__init__.py
+++ b/bootstrapvz/providers/gce/__init__.py
@@ -16,8 +16,8 @@ def validate_manifest(data, validator, error):
 
 
 def resolve_tasks(taskset, manifest):
+    from bootstrapvz.common.releases import stretch
     taskset.update(task_groups.get_standard_groups(manifest))
-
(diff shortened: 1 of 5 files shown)


bootstrapvz/providers/gce/__init__.py
```diff
@@ -42,6 +42,13 @@ def resolve_tasks(taskset, manifest):
                     ])
     taskset.discard(grub.SetGrubConsolOutputDeviceToSerial)
 
+    # Temporary fix for Stretch
+    if manifest.release == testing:
```
andsens: You specify`stretch` in the comment, so you should use `boostrapvz.common.releases.stretch`. This ensures forwards compatibility once stretch becomes `stable` (regardless of how temporary this is).

zmarano: Sounds good to me, done!

andsens:
Great stuff 👍 
Merged.

PR #358, opened 2017-01-19
Fixing device path in the gce expand-root

### LINKED ISSUE #304: expand-root init.d script is not assigning root_device_path correctly
With the changes to the common expand-root script, the device is filled in with info.volume.device_path which turns out to point to the device of the install volume. If you are using a raw volume for example this means you end up with /dev/loop0 as the resulting device that is supposed to be expanded. This is obviously wrong and should be whatever the boot drive actually is (normally /dev/sda but could be other things).

eranmeir:
fixes #304

### DIFF
--- a/bootstrapvz/providers/gce/tasks/initd.py
+++ b/bootstrapvz/providers/gce/tasks/initd.py
@@ -13,4 +13,4 @@ class AdjustExpandRootDev(Task):
     def run(cls, info):
         from bootstrapvz.common.tools import sed_i
         script = os.path.join(info.root, 'etc/init.d/expand-root')
-        sed_i(script, '/dev/loop0', '/dev/sda')
+        sed_i(script, info.volume.device_path, '/dev/sda')


andsens:
I'm sorry, but I don't understand this PR at all.  
Why is there a custom `AdjustExpandRootDev` task for gce at all? Why isn't [`common.tasks.AdjustExpandRootScript`](https://github.com/andsens/bootstrap-vz/blob/6866adf734c5436ed86da699f497ee933cedae17/bootstrapvz/common/tasks/initd.py#L60) used. I assume the common [`expand-root`](https://github.com/andsens/bootstrap-vz/blob/6866adf734c5436ed86da699f497ee933cedae17/bootstrapvz/common/assets/init.d/expand-root) script is used. That one has the root device hardcoded to `/dev/xvda`, I don't see how `/dev/loop0` or `info.volume.device_path` factor into this at all.

eranmeir:
Disclaimer: I haven't used bootstrap-vz extensively, nor claim to understand the code perfectly. This is an attempt to fix an error I encountered (described below).

It seems to me that `common.tasks.AdjustExpandRootScript` replaces `/dev/xvda` with (formatted) `info.volume.device_path`. PR #303 added an `AdjustExpandRootDev` task  to the gce provider. This task again updates the root device with `/dev/sda`. But it has a hard-coded pattern to match , I assume under the false assumption the volume is always `/dev/loop0`.

As commented in #304, `expand-root` would then give the following error when running: 
```
Adjusting the expand-root device
There were 0 instead of 1 matches for the expression `/dev/loop0' in the file `/target/3bdaa011/root/etc/init.d/expand-root'
```

The effect is that each provider (or plugin?) can adjust the expanded device using a custom task that follows `initd.AdjustExpandRootScript`. This effectively fixes #304 - although a nicer way would be to add this as a configuration option.

zmarano:
Actually, if you want to hold off doing this for a little bit - I am creating a plugin for expand-root as opposed to the current built-in version (to resolve some problems with Stretch). The built-in version has a number of problems including this issue. I plan on leaving the existing built-in version in place for the time being. And, if you want to use the simplified plugin version you can (GCE will).

zmarano:
FYI: See https://github.com/andsens/bootstrap-vz/pull/360

PR #360, opened 2017-01-25
Create a plugin for expand_root to replace the internal common task.

### LINKED ISSUE #304: expand-root init.d script is not assigning root_device_path correctly

zmarano:
Simplify how expand_root works for Jessie and Stretch. See the README and GCE use case for how to use the plugin. Switch GCE jessie and stretch builds to use this plugin. Also fixes #304


andsens:
Looks awesome 👍   
Merging...

PR #366, opened 2017-03-07
GCE: Add the file package which many tests expect to exist. It seems to have disappeared in some recent upstream dependency change.

zmarano:


andsens:
Merging...  
Thanks Zach :-)

PR #367, opened 2017-03-15
EC2: Update the URL to dowload the ixgbevf driver

grim-luminal:
The URL for the `ixgbevf` driver is currently returning a 404. Updating
to the new mirror URL gets EC2 builds with enhanced networking support
working again.


andsens:
👍

PR #371, opened 2017-03-28
Fix bug in Stretch builds, sshd_config no longer contains PermitRootLogin.

zmarano:
See bug for context: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=852781


andsens:
Shouldn't it be `sed_i(sshdconfig_path, '^#?PermitRootLogin .*', 'PermitRootLogin no')`?  
This way we also accommodate the commented out version.

zmarano:
Oh yeah- that's way better. Changed.

andsens:
Great stuff :-)  
Merging...

PR #372, opened 2017-04-14
Fix an ordering issue with the expand-root plugin.

zmarano:
For whatever reason, there was an ordering issue where this class was being run before anything else and failing because there wasn't a jessie-backports sources list yet (not to mention a filesytem at all). Moving this to the package_installation phase fixes the problem.

### DIFF
--- a/bootstrapvz/plugins/expand_root/tasks.py
+++ b/bootstrapvz/plugins/expand_root/tasks.py
@@ -1,7 +1,7 @@
 from bootstrapvz.base import Task
 from bootstrapvz.common import phases
+from bootstrapvz.common.tasks import apt
 from bootstrapvz.common.tasks import initd
-from bootstrapvz.common.tasks import packages
 from bootstrapvz.common.tools import log_check_call
 from bootstrapvz.common.tools import rel_path
 from bootstrapvz.common.tools import sed_i
@@ -14,7 +14,7 @@ ASSETS_DIR = rel_path(__file__, 'assets')
 class InstallGrowpart(Task):
     description = 'Adding necessary packages for growpart.'
     phase = phases.preparation
-    successors = [packages.AddManifestPackages]
+    predecessors = [apt.AddBackports]
 
     @classmethod
     def run(cls, info):


andsens:
> there was an ordering issue where this class was being run before anything else and failing because there wasn't a jessie-backports sources list yet

I see. We should keep the task in the preparation phase though, just add the [`apt.AddBackports`](https://github.com/andsens/bootstrap-vz/blob/96ebfe11fad2b18e1a63e496e2daf708348edc5b/bootstrapvz/common/tasks/apt.py#L73) task as a successor. Right now you can risk the task being run after `InstallPackages`(https://github.com/andsens/bootstrap-vz/blob/96ebfe11fad2b18e1a63e496e2daf708348edc5b/bootstrapvz/common/tasks/packages.py#L24).

> not to mention a filesytem at all

Doesn't matter, all package adding and so on is virtual, it's just a `set()` that we are updating, which `InstallPackages` uses later on.

Packages should be added in the preparation phase, though I can see a few are actually placed in the pkg installation phase, which is wrong.

zmarano:
Ah gotcha... predecessors = [apt.AddBackports] works. Is there any way to validate ordering when testing things out? --debug doesn't show you much except that in this case we saw that this task was being run way too early.

andsens:
> Is there any way to validate ordering when testing things out?

Yes, I have actually created an overview with d3.js to visualize task groupings etc., but I haven't worked on it in a while. The number of tasks has grown significantly, rendering the graphic a little unwieldy. The task names are only shown on mouseover which is a little useless. Tasks are coloured by the module they belong to.

You can see it here: https://bootstrap-vz.readthedocs.io/en/master/developers/taskoverview.html
It can be [generated](https://github.com/andsens/bootstrap-vz/blob/a40c639c3c56c1d6c82162ba7130cc92c7ed403a/docs/taskoverview.py) locally with `tox -e docs` and then served with `(cd docs/_build/html; python -m SimpleHTTPServer 8080)`. I hope that helps :-)

Merging...

Issue #376, opened 2017-05-02
Stretch/kvm fails with  for lack of generate-ssh-hostkeys


andrewbogott:
I ran into this with a custom manifest, but using the built in stretch-virtio-partitions.yml manifest I still can't build an image.  The build errors out with this traceback:

Traceback (most recent call last):
  File "/usr/local/bin/bootstrap-vz", line 9, in <module>
    load_entry_point('bootstrap-vz==0.9.11', 'console_scripts', 'bootstrap-vz')()
  File "/usr/local/lib/python2.7/dist-packages/bootstrap_vz-0.9.11-py2.7.egg/bootstrapvz/base/main.py", line 29, in main
    dry_run=opts['--dry-run'])
  File "/usr/local/lib/python2.7/dist-packages/bootstrap_vz-0.9.11-py2.7.egg/bootstrapvz/base/main.py", line 111, in run
    tasklist.run(info=bootstrap_info, dry_run=dry_run)
  File "/usr/local/lib/python2.7/dist-packages/bootstrap_vz-0.9.11-py2.7.egg/bootstrapvz/base/tasklist.py", line 44, in run
    task.run(info)
  File "/usr/local/lib/python2.7/dist-packages/bootstrap_vz-0.9.11-py2.7.egg/bootstrapvz/common/tasks/initd.py", line 22, in run
    copy(src, dst)
  File "/usr/lib/python2.7/shutil.py", line 119, in copy
    copyfile(src, dst)
  File "/usr/lib/python2.7/shutil.py", line 82, in copyfile
    with open(src, 'rb') as fsrc:
IOError: [Errno 2] No such file or directory: '/usr/local/lib/python2.7/dist-packages/bootstrap_vz-0.9.11-py2.7.egg/bootstrapvz/common/assets/init.d/generate-ssh-hostkeys'

andrewbogott:
Indeed, common/assets doesn't seem to get installed at all by setup.py.

andsens:
There, that should do it :-)

PR #377, opened 2017-05-02
MANIFEST.in:  Include a few directories that we really need installed.

andrewbogott:
There are probably more dirs missing but these two scratch my itch.

Fixes: #376

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,5 +1,7 @@
 include LICENSE
 include manifests/*
 recursive-include bootstrapvz assets/*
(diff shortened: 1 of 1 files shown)


andsens:
Quick update for all PRs: I've seen them and will get to them. I'm a little pre-occupied at the moment, so it might be two-three weeks before I answer. (and thanks for the PR btw :-))

andsens:
Are you sure that's the right syntax? All `recursive-include` arguments after the first are patterns, not paths.

andsens:
Ah! Figured it out. I added the changes myself :-)

PR #378, opened 2017-05-14
Fix loopdevice partitions not being unmapped correctly

Exy13:
Removes undeleted loopdevice partitions created by kpartx.
This was needed as "bootstrap-vz tries very hard to clean up after itself " and because LVM tries to read from these files and gives off warnings.
Since Util-linux 2.21, losetup is able to scan the partition table of the newly created loop device. Therefore it is able to unmap them as well. By adding a --partscan, losetup tries to map the partitions when the loop device is mounted, and will unmap them when it is detached.

### DIFF
--- a/bootstrapvz/common/fs/loopbackvolume.py
+++ b/bootstrapvz/common/fs/loopbackvolume.py
@@ -15,7 +15,7 @@ class LoopbackVolume(Volume):
         log_check_call(['truncate', size_opt, self.image_path])
 
     def _before_attach(self, e):
-        [self.loop_device_path] = log_check_call(['losetup', '--show', '--find', self.image_path])
+        [self.loop_device_path] = log_check_call(['losetup', '--show', '--find', '--partscan', self.image_path])
         self.device_path = self.loop_device_path
 
     def _before_detach(self, e):


andsens:
Quick update for all PRs: I've seen them and will get to them. I'm a little pre-occupied at the moment, so it might be two-three weeks before I answer. (and thanks for the PR btw :-))

andsens:
> Removes undeleted loopdevice partitions created by kpartx.

Wouldn't this be handled by [`kpartx -d`](https://github.com/andsens/bootstrap-vz/blob/9ac970b715698bc193ca08354ef109ac4d80f08c/bootstrapvz/base/fs/partitionmaps/abstract.py#L119)?

Exy13:
It should be expected, however kpartx leaves 'ghost' files such as /dev/loop?p?.
These files don't serve any purpose don't appear to be mapped to anything.
And they cause input/output error when accessed.

andsens:
Ah, understood. I assume adding `--partscan` does not clash with `kpartx -d` in any way, it's just a no-op if all is good, correct?

Exy13:
Indeed, when losetup --partscan is launched, it will search for a partition table on a blank loop device, therefore not doing anything.

andsens:
ok cool, merging :-)

PR #379, opened 2017-05-14
Complete PATH for distributions that only have "/usr/bin" in PATH

Exy13:
Bootstrap-vz expects to have a path that contains /bin, /sbin and
/usr/sbin, but some distributions (Archlinux for example)
have all these directories merged into /usr/bin.

### DIFF
--- a/bootstrapvz/base/main.py
+++ b/bootstrapvz/base/main.py
@@ -22,6 +22,10 @@ def main():
     from manifest import Manifest
     manifest = Manifest(path=opts['MANIFEST'])
 
+    # Set-up Architecture independent path
+    if "usr/sbin" not in os.environ["PATH"]:
+        os.environ["PATH"] += os.pathsep + os.pathsep.join(["/usr/sbin", "/sbin", "/bin"])
+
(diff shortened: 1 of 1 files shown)


andsens:
This doesn't make sense. bootstrap-vz doesn't mess with `$PATH`, the whole point is that it inherits that from the parent process, which in most cases is the users shell. Are you sure that your path is properly set up? Because that might be the root cause for the problem.

Closing for now, feel free to reopen if I'm wrong :-)

GabrielL:
I have the same issue. On Archlinux, the default PATH is: "/usr/local/sbin:/usr/local/bin:/usr/bin", because all bin directories are merged into /usr/bin (with symlinks to /bin… for compatibility). As bootstrap-vz is installing debian systems, all chroot commands will expect a PATH that will looks like the one from debian.

andsens:
ah, so let me get this straight. The actual problem is when commands are run inside the chroot, because the PATH from the host is inherited? In that case I think we need a more sophisticated solution where PATH is only set when using chroot.

GabrielL:
Yes this would be better.

PR #381, opened 2017-05-26
Add LVM as a disk backend

Exy13:
Enables the use of Logical Volumes as disk backends.
This was needed as we store our VMs through LVM and wanted to build 
the image directly onto a lv.

It uses an existing volume group and has no support for creating a new one.
It will not override an existing logical volume and fail gracefully.

The lv is created, activated and then mounted as a loop device.
The boostraping process is then launched on the loop device.
Once the process is completed, the lv is unmounted and desactivated.

The created lv will be deleted should the boostraping process fail.

The lv must be activated before use.

A manifest has been included for testing purposes.

### DIFF
--- a/bootstrapvz/base/manifest-schema.yml
+++ b/bootstrapvz/base/manifest-schema.yml
@@ -132,6 +132,12 @@ properties:
     additionalProperties: false
   volume:
     type: object
+    oneOf:
+        - $ref: '#/definitions/standardvolume'
+        - $ref: '#/definitions/logicalvolume'
+definitions:
+  standardvolume:
+    type: object
     properties:
       backing: {type: string}
       partitions:
@@ -141,7 +147,20 @@ properties:
           - $ref: '#/definitions/partition_table'
     required: [partitions]
     additionalProperties: false
-definitions:
+  logicalvolume:
+    type: object
+    properties:
+      backing:
+        enum: [lvm]
+      volumegroup : {type: string}
+      logicalvolume: {type: string}
+      partitions:
+        type: object
+        oneOf:
+          - $ref: '#/definitions/no_partitions'
+          - $ref: '#/definitions/partition_table'
+    required: [partitions]
+    additionalProperties: false
   absolute_path:
     type: string
     pattern: ^/[^\0]+$
--- a/bootstrapvz/providers/kvm/__init__.py
+++ b/bootstrapvz/providers/kvm/__init__.py
@@ -1,6 +1,6 @@
 from bootstrapvz.common import task_groups
 import tasks.packages
-from bootstrapvz.common.tasks import image, loopback, initd, ssh
+from bootstrapvz.common.tasks import image, loopback, initd, ssh, logicalvolume
 
 
 def validate_manifest(data, validator, error):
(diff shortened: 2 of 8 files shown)


manifests/examples/kvm/jessie-lvm.yml
```diff
@@ -0,0 +1,27 @@
...
+system:
+  release: jessie
+  architecture: amd64
+  bootloader: grub
+  charmap: UTF-8
+  locale: en_US
+  timezone: UTC
+volume:
+  backing: lv
+  logicalvolume:
```
andsens: I don't think there's any need to put `vg` and `lv` under another key. Just spell them out directly in `volume`. So: `volume: {logivalvolume: "lvname", volumegroup: "vgname"}`

bootstrapvz/base/manifest-schema.yml
```diff
@@ -134,6 +134,7 @@ properties:
     type: object
     properties:
       backing: {type: string}
+      logicalvolume: {type: object}
```
andsens: For better manifest validation the volume object should be split into a `standardvolume` definition and `logicalvolume` definition, like so:
```
  volume:
    type: object
    oneOf:
      - $ref: '#/definitions/standardvolume'
      - $ref: '#/definitions/logicalvolume'
definitions:
  ...
  standardvolume:
    type: object
    properties:
      backing: {type: string}
      partitions:
        type: object
        oneOf:
          - $ref: '#/definitions/no_partitions'
          - $ref: '#/definitions/partition_table'
    required: [partitions]
    additionalProperties: false
  logicalvolume:
    type: object
    properties:
      backing: lvm
      volumegroup: {type: string}
      logicalvolume: {type: string}
      partitions:
        type: object
        oneOf:
          - $ref: '#/definitions/no_partitions'
          - $ref: '#/definitions/partition_table'
    required: [volumegroup, logicalvolume, partitions]
    additionalProperties: false
```

bootstrapvz/providers/kvm/__init__.py
```diff
@@ -12,14 +12,20 @@ def resolve_tasks(taskset, manifest):
...
-                    loopback.AddRequiredCommands,
-                    loopback.Create,
                     initd.InstallInitScripts,
                     ssh.AddOpenSSHPackage,
                     ssh.ShredHostkeys,
                     ssh.AddSSHKeyGeneration,
-                    image.MoveImage,
                     ])
+    if manifest.volume.get('logicalvolume', []):
+        taskset.update([logicalvolume.AddRequiredCommands,
```
Exy13: Which providers do you have in mind ?

Exy13:
I have updated the various issues.

bootstrapvz/providers/kvm/__init__.py
andsens: Come to think of it, there might not be any. This is pretty kvm specific, since the volume is kind of tied to the host machine, correct?  
I'm ok with merging this if you squash your commits :-)

Exy13: The commits have been squashed.

Issue #380, opened 2017-05-16
"No such file or directory" after mapping volume permissions


RX14:
```
bootstrap-vz/bootstrap-vz bootstrap-vz/manifests/official/ec2/ebs-jessie-amd64-hvm.yml
Generating tasklist
Adding commands required for bootstrapping Debian
Determining required external commands for EC2 bootstrapping
Adding commands required for formatting
Adding commands required for partitioning the volume
Checking availability of external commands
Adding openssh package
Creating workspace
Adding locale package to bootstrap installation
Initializing grub standard configuration
Adding default release sources
Adding backports to the apt sources
Adding cloud-init package and sudo
Retrieving instance metadata
Adding grub package
Getting AWS credentials
Connecting to EC2
Found credentials in environment variables.
Silence boto debug logging
Determining the AMI name
Starting new HTTPS connection (1): ec2.us-east-1.amazonaws.com
Adding DKMS and kernel header packages
Adding packages from the manifest
Adding image packages required for EC2
Adding preferences from the manifest
Creating the EBS volume
Resetting dropped connection: ec2.us-east-1.amazonaws.com
Attaching the volume
Resetting dropped connection: ec2.us-east-1.amazonaws.com
Partitioning the volume
Mapping volume partitions
[Errno 2] No such file or directory
Traceback (most recent call last):
  File "/root/bootstrap-vz/bootstrapvz/base/main.py", line 111, in run
    tasklist.run(info=bootstrap_info, dry_run=dry_run)
  File "/root/bootstrap-vz/bootstrapvz/base/tasklist.py", line 44, in run
    task.run(info)
  File "/root/bootstrap-vz/bootstrapvz/common/tasks/partitioning.py", line 38, in run
    info.volume.partition_map.map(info.volume)
  File "/root/bootstrap-vz/bootstrapvz/base/fs/partitionmaps/abstract.py", line 61, in map
    self.fsm.map(volume=volume)
  File "/usr/local/lib/python2.7/dist-packages/fysom/__init__.py", line 286, in fn
    if self._before_event(e) is False:
  File "/usr/local/lib/python2.7/dist-packages/fysom/__init__.py", line 321, in _before_event
    return getattr(self, fnname)(e)
  File "/usr/local/lib/python2.7/dist-packages/fysom/__init__.py", line 89, in _callback
    return func(obj, *args, **kwargs)
[...]

PR #382, opened 2017-05-31
Fix unfailing CheckExternalCommands

Exy13:
On Unix, with `shell=True`, the shell default to /bin/sh.
Using `Popen(['type', command], shell=True)` is equivalent to calling
`Popen(['/bin/sh', '-c', 'type', command])`.
In this case `'command'` becomes a positional parameter to the shell,
and not an argument to the command `'type'`.

One solution would be to pass a single string as parameter.
The problem is that with `shell=True`, we wouldn't be safe from a shell injection,
so it is wiser to use a python only solution.

The package distutils is part of the standard distribution, so it doesn't add
extra dependencies.
The method find_executable has the same behaviour as 'which' on bash, it 
will return the path to the executable or `None` if it was not found.

Logging needs to be added as it was previously handled by log_check_call().

This fixes [#380](https://github.com/andsens/bootstrap-vz/issues/380) which was broken [here](https://github.com/andsens/bootstrap-vz/commit/bb41ba68dd70ef20518fbe7969accaff93c22300).

### DIFF
--- a/bootstrapvz/common/tasks/host.py
+++ b/bootstrapvz/common/tasks/host.py
@@ -9,14 +9,14 @@ class CheckExternalCommands(Task):
 
     @classmethod
     def run(cls, info):
-        from ..tools import log_check_call
-        from subprocess import CalledProcessError
         import re
+        import logging
+        from distutils.spawn import find_executable
         missing_packages = []
+        log = logging.getLogger(__name__)
         for command, package in info.host_dependencies.items():
-            try:
-                log_check_call(['type', command], shell=True)
-            except CalledProcessError:
+            log.debug('Checking availability of ' + command)
+            if find_executable(command) is None:
                 if re.match('^https?:\/\/', package):
                     msg = ('The command `{command}\' is not available, '
                            'you can download the software at `{package}\'.'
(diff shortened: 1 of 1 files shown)


andsens:
I checked out the implementation of `find_executable`, and it honestly leaves a little to be desired:
```

def find_executable(executable, path=None):
    """Tries to find 'executable' in the directories listed in 'path'.

    A string listing directories separated by 'os.pathsep'; defaults to
    os.environ['PATH'].  Returns the complete filename or None if not found.
    """
    if path is None:
        path = os.environ['PATH']
    paths = path.split(os.pathsep)
    base, ext = os.path.splitext(executable)

    if (sys.platform == 'win32' or os.name == 'os2') and (ext != '.exe'):
        executable = executable + '.exe'

    if not os.path.isfile(executable):
        for p in paths:
            f = os.path.join(p, executable)
            if os.path.isfile(f):
                # the file exists, we have a shot at spawn working
                return f
        return None
    else:
        return executable
```

Note that there is no check for executability. Also, we will not be able to check the availability of shell functions or shell builtins with this.

I think your suggestion of making the whole checking command a string might be a better solution to this problem, we needn't worry about shell injection too much, since all of the input is trusted.

Exy13:
I have updated with an executability check.
Shell builtins are not used in the project, and I see no valid use for them.
The same goes for shell functions, wouldn't it better to implement them in python directly ?

andsens:
eh, true. Alright, I'm fine with merging this. Nice touch with adding debug logging, since there is no `log_check_call` any longer :-)

PR #383, opened 2017-06-16
Add qcow2 as a disk backend for KVM

Exy13:
This enables building an image with qcow2 as a disk backend.
Qcow2 is only supported by KVM.

The update implements a new class that inherits from QEMUVolume and just
needs the extension and file format.

A manifest has been included for testing purposes.


andsens:
Hah, seems like I did something right with the volume backing :-)

Merging...

PR #384, opened 2017-06-16
Promote stretch to stable, add buster as testing, jessie to oldstable.

zmarano:
Stretch release to stable is June 17(tonight/tomorrow depending on time zone).

### DIFF
--- a/bootstrapvz/common/releases.py
+++ b/bootstrapvz/common/releases.py
@@ -31,7 +31,8 @@ class _ReleaseAlias(_Release):
         return self.alias
 
 
-sid = _Release('sid', 10)
+sid = _Release('sid', 11)
+buster = _Release('buster', 10)
 stretch = _Release('stretch', 9)
 jessie = _Release('jessie', 8)
 wheezy = _Release('wheezy', 7)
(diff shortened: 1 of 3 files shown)


andsens:
Woops, tests failing. These need fixing: [tests/unit/releases_tests.py](https://github.com/andsens/bootstrap-vz/blob/c23e73c2223a9e4f25615078b384b2e32b5f141d/tests/unit/releases_tests.py#L25-L34)

zmarano:
Oh yeah, tests ;) Fixed!

andsens:
Nice 👍 
I'll fix these myself :-)
```
def test_not_identity():
    assert releases.wheezy is not releases.stable
    assert releases.stable is not releases.jessie
```
And maybe document the idea behind them :-)

zmarano:
Yeah I saw those- I don't really understand why those work because ```assert releases.stable is not releases.jessie``` should have failed before.

andsens:
It shouldn't. It's an identity check `==` is not the same as `is`.  
So `stable == jessie` and `stable is stable` but `stable is not jessie`. Not sure if that clears it up or adds confusion.

zmarano:
I see what you're saying- but yeah its still really confusing :)

PR #385, opened 2017-06-24
Fix lone minimize_size/dpkg without apt

liori:
The `filter_tasks` variable is only used in the `if 'dpkg'` branch, but curiously it is only initialized in the `if 'apt'` branch.


andsens:
Woops, you're right. Merging.

PR #387, opened 2017-07-09
minimize_size: New `repack` option for qemu-img-based repacking

liori:
I haven't actually tested this on vdi, but based on `qemu-img`'s documentation, it should work just as well.

### DIFF
--- a/bootstrapvz/plugins/minimize_size/README.rst
+++ b/bootstrapvz/plugins/minimize_size/README.rst
@@ -15,16 +15,20 @@ footprint:
    deallocate unused sectors on the volume. On an unpartitioned volume
    this will be done for the entire volume, while it will only happen on
    the root partition for partitioned volumes.
--  Use
-   `vmware-vdiskmanager <https://www.vmware.com/support/ws45/doc/disks_vdiskmanager_eg_ws.html>`__
-   to shrink the real volume size (only applicable when using vmdk
-   backing). The tool is part of the `VMWare
-   Workstation <https://my.vmware.com/web/vmware/info/slug/desktop_end_user_computing/vmware_workstation/10_0>`__
-   package.
+-  Shrink the real volume size. Supported tools are:
+
+   -  `vmware-vdiskmanager <https://www.vmware.com/support/ws45/doc/disks_vdiskmanager_eg_ws.html>`__
+      (only applicable when using vmdk backing). The tool is part of the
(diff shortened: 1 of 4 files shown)


andsens:
Cool! Would it ever make sense to both use vmware-diskmanager *and* qemu-img? Otherwise I'd suggest making `shrink` an enum with the two options and `True/False` as backwards compat.

liori:
I don't know, I don't have VMWare installed anywhere. I guess there should be little benefit in running them both. I'll make it into an enum then. Do you prefer squashed merges or should I add the changes as a separate commit?

andsens:
Squashed would be best :-)

andsens:
Any update on this?

liori:
Sorry for the delay. The change is now implemented. Thinking about it, I'd make a small refactoring here and split `shrink.AddRequiredCommands` into three separate tasks for `zerofree` and both types of `shrink` — though in a separate commit or push request, later. Is that desirable?

bootstrapvz/plugins/minimize_size/README.rst
```diff
@@ -35,8 +39,13 @@ Settings
    Default: false
    ``optional``
 -  ``shrink``: Whether the volume should be shrunk. This setting works
-   best in conjunction with the zerofree tool.
-   Valid values: true, false
+   best in conjunction with the zerofree tool. Valid values:
+
+   -  false: Do not shrink.
+   -  ``vmware-vdiskmanage`` or true: Shrink using the `vmware-vdiskmanager`
```
andsens: Woops, typo. Missing `r` in `vmware-vdiskmanager`

bootstrapvz/plugins/minimize_size/tasks/shrink.py
```diff
@@ -8,6 +8,20 @@
 import os
 
 
+def get_shrink_type(plugins):
```
andsens: The tasks module should really only contain actual `Task`s. I recommend putting this in `__init__`

andsens:
A split-up of the required commands task would be appreciated. Then you don't need the helper function either, you can just inline it in `resolve_tasks()`.

liori:
I fixed the typo and added a commit that implements the split of `shrink.AddRequiredCommands`. The helper function is still also used in the `validate_manifest` function, so I didn't inline it; but I moved it to the `minimize_size/__init__.py` file. I'm not sure if the identifiers/description strings for new tasks are proper?

andsens:
Looks great! Merging.

The past tense of "shrink" is "shrunk" btw., not that it matters, it's a temp file after all :-)

Issue #389, opened 2017-08-03
escaping of {foo}: possible?


indigoid:
I wanted to insert the verbatim text `${USER}` into a file via a `sed` command in the `commands` section in my manifest, without any expansion taking place.

But I don't seem to be able to escape the `{}`, and get `KeyError: USER` in the log output. Is this escaping possible? I have tried escaping with backslashes, and this did not have the desired effect.

NB: am invoking `bootstrap-vz` via `kube-deploy/imagebuilder`, but have confirmed that the Golang templating used there is not interfering. The final generated `template.yaml` contains this line:

`- [ 'chroot', '{root}', 'sed', '-i', '1i auth [success=done new_authtok_reqd=done default=die] pam_google_authenticator.so nullok user=totp secret=/efs/totp/${USER}/.google_authenticator', '/etc/pam.d/sshd' ]`

andsens:
This is from memory, but you should be able to just use `${{USER}}`, thats the python string template way of doing things. This should probably be documented in [manifests/README.rst](https://github.com/andsens/bootstrap-vz/blob/9364b18288251abeac5e0a285694fd915dce731e/manifests/README.rst) somewhere.

indigoid:
it certainly is. thanks!

PR #390, opened 2017-08-03
brief note about escaping braces in manifests

jsleeio:
as mentioned in #389 ... `{{foo}}` is indeed the thing to do 👍

### DIFF
--- a/manifests/README.rst
+++ b/manifests/README.rst
@@ -21,6 +21,10 @@ setting. Not all settings support this though, to see whether embedding
 a manifest variable in a setting is possible, look for the
 ``manifest vars`` label.
 
+To insert a literal ``{foo}``, eg. in a shell command where you may want
+to use the expression ``${foo}``, use double braces instead, ie.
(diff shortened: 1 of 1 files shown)


manifests/README.rst
```diff
@@ -21,6 +21,10 @@ setting. Not all settings support this though, to see whether embedding
 a manifest variable in a setting is possible, look for the
 ``manifest vars`` label.
 
+To insert a literal ``{foo}``, eg. in a shell command where you may want
+to use the expression ``${foo}``, use double braces instead, ie.
+``{{foo}}``.
```
andsens: Nice, though the sentence structure might be a little confusing, how about:
```
To insert a literal ``{foo}`` use double braces, that is ``{{foo}}``.
For example in a shell command where you may want to use the
expression ``${foo}``, use ``${{foo}}`` instead.

andsens:
Nice. Thank you @indigoid :-)

PR #392, opened 2017-08-23
Don’t crash on empty AMI name

gplessis:
Allow empty names for existing AMIs


andsens:
Looks good, thanks @gplessis!

PR #393, opened 2017-08-24
Update Google Compute Engine package and repository structure.

zmarano:
GCE packages are now Debian policy compliant and therefore we have new package names and repositories.


andsens:
LGTM, merging :-)

PR #394, opened 2017-08-25
Instance store AMIs with HVM virtualization

gplessis:
I've been able to:

1. build instance store AMIs with HVM virtualization
2. launch instances successfully
3. test them without any issue

just by allowing such a setup in the code and by using these two manifests.

If you plan to support such a (instance type / virtualization) couple in the future, feel free to review and decide the fate of this PR 😬

### DIFF
--- a/bootstrapvz/providers/ec2/__init__.py
+++ b/bootstrapvz/providers/ec2/__init__.py
@@ -31,18 +31,10 @@ def validate_manifest(data, validator, error):
 
     bootloader = data['system']['bootloader']
     virtualization = data['provider']['virtualization']
-    backing = data['volume']['backing']
-    partition_type = data['volume']['partitions']['type']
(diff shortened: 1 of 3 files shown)


andsens:
Sweet. At the time I coded this the boto version and AWS support may have prohibited this. I will have to test it out in the weekend :-)

andsens:
Wait. wth github? I had a review pending I didn't know about. Sorry @gplessis.  
I hope this is still relevant: Are you able to partition the root volume when using S3 backed instances? Because the code you removed also disallows that.

gplessis:
No pb :)

I haven't tried to have multiple partitions on the root volume, but I ended up having only one - `/dev/xvda1` - mounted as `/`.

andsens:
OK. So I *think* you need to keep the partitioning guard in there. Last time I tested, it wasn't possible. The xvda**1** suggests that this is still the case.

PR #395, opened 2017-08-31
Add GCE buster builds.

zmarano:
- And generally support buster for test builds.


andsens:
Looks good! Merging...

PR #396, opened 2017-09-03
Stretch update commit

NeatNerdPrime:
* updated module to allow for puppetlabs PC1 release install in
  Debian Stretch. Tox is green.


andsens:
Nice job! Merging.

p.s.: Isn't the keyring the same for all versions? Do you need to have a separate for each release?

NeatNerdPrime:
Hi @andsens , thanks for your compliment.

Indeed it seems to be an identical keyring for each release. However,  since you cannot ass-u-me that would always be the case, i opted for keeping them separate, in case puppetlabs would decide to use a different keyring for a different release in the future. It also makes each release consistent. and for those few bytes more... meh.

If you would like i can point them to the single file, removing separation for each release.

andsens:
Nah, it's fine. Though if they were to change the key I'd wager they would change it for the entire repo, meaning you'd have to update all the keys regardless. Don't worry though, this is fine for now :-)

Issue #397, opened 2017-09-14
Installing enhanced networking kernel driver using DKMS fails in stretch


jmporcelg:
Building debian stretch ami with Jenkins and bootstrap-vz from today I found DKMS fails to build the driver. Not sure is my manifest.yml file or something else.

However, changing [the tarball version ixgbevf-3.2.2 here](https://github.com/andsens/bootstrap-vz/blob/a6200ff629f83414040e3605569004a2f1eb24d6/bootstrapvz/providers/ec2/tasks/network.py#L91) for this one https://downloadmirror.intel.com/18700/eng/ixgbevf-4.2.1.tar.gz fixes the issue and the build completes.

andsens:
Hm, what does the log say? Regardless, we should just update the driver. Could you maybe send a PR with the fix? Maybe check if the release is greater than stretch and then just use the new URL, keep it the same for anything below, just for backwards compat.

jmporcelg:
Agreed, best approach is to update the driver regardless. Working code in the above PR.

andsens:
Merged :-)

PR #398, opened 2017-09-15
#397 fix: Installing networking kernel driver using DKMS fails in stretch

jmporcelg:
Apply new version when release is stretch or higher.


andsens:
wth? I thought I merged that ~ a week ago. Sorry about the delay.  
Good job @jmporcelg!

PR #400, opened 2017-10-02
Update misleading documentation

Exy13:
Commands passed to the commands plugin are not chrooted in the image.
Therefore the given example is misleading as it will remove the locales on the host and not on the image.

The use of `chroot {root}` should be advised.


andsens:
good catch. Merging.
Additionally, the arguments should be separated in the list, otherwise you get even weirder results if your root is on a path with spaces.

PR #407, opened 2017-11-09
Remove the bits that disabled IPv6 for GCE.

zmarano:


zmarano:
The travis failures have nothing to do with the changes I made in this PR- not sure how you want to fix that but it should probably be in a separate PR.

andsens:
Cool beans! Merging...

> The travis failures have nothing to do with the changes I made in this PR- not sure how you want to fix that but it should probably be in a separate PR.

hm, I should probably pin flake8 to a specific version. New pep8 guidelines are sure to break stuff that wasn't broken before.

PR #412, opened 2017-11-22
Prevent guest_additions ValueError

CMeza99:
If a second `:` is a line from the return from apt-cache,  `key, value = line.split(':')` will result in an exception ValueError. Setting the _maxsplit_ parameter to 1 will prevent this.


andsens:
Actually this task shouldn't determine that kernel version at all. There is a specific task for that: https://github.com/andsens/bootstrap-vz/blob/9445273b8e7ea55d4766b45152bcda3c15911138/bootstrapvz/common/tasks/kernel.py#L29
Would you mind changing it to just use `info.kernel_version`?

CMeza99:
will do

PR #414, opened 2017-11-23
allow networking task when netbase pkg is present

CMeza99:
Currently if the variant minbase is used, the network task is skipped. But if the netbase package is installed, the network should be configured.


andsens:
LGTM 👍

PR #415, opened 2017-11-28
Added security field to packages.

ashegedus:
Added a security mirror setting to packages.


andsens:
Nice work! You need to add some manifest documentation though: https://github.com/andsens/bootstrap-vz/blob/9445273b8e7ea55d4766b45152bcda3c15911138/manifests/README.rst  
You could also add the instruction to one of the example manifests so that it is part of the dry run integration tests.

ashegedus:
Updated README.rst.

andsens:
Nice job! Merging.

PR #417, opened 2017-11-29
Add password-crypt to root password plugin

ashegedus:
Added an option to the root_password plugin to use a password hash.

- Added password-crypted to root_password plugin.
- Updated the docs for root_password plugin
- Added password-crypted to example manifest for stretch vagrant


andsens:
Great work, especially for keeping backwards compat. with existing manifests :-)

PR #422, opened 2017-12-20
update intel ixgbevf url

CMeza99:
current url is a 404

there is also https://downloadmirror.intel.com/18700/eng/ixgbevf-4.3.2.tar.gz


andsens:
👍

PR #426, opened 2018-01-08
Remove ED25519 ssh host key pair from image

mengelmann:
ED25519 ssh host key pair is not removed from ```/etc/ssh```:

```
$ ls -l /etc/ssh/ssh_host_*
-rw------- 1 root root 399 Jan  8 16:39 etc/ssh/ssh_host_ed25519_key
-rw-r--r-- 1 root root  94 Jan  8 16:39 etc/ssh/ssh_host_ed25519_key.pub
```

ED25519 ssh key pair is created since OpenSSH 6.5 (6.7 in Debian jessie) https://wiki.debian.org/SSH#Installation_of_the_server

### DIFF
--- a/bootstrapvz/common/assets/init.d/jessie/generate-ssh-hostkeys
+++ b/bootstrapvz/common/assets/init.d/jessie/generate-ssh-hostkeys
@@ -5,7 +5,7 @@
 # Required-Stop:  
 # Should-Start:   
 # Should-Stop:    
-# Default-Start:  2 3 4 5
+# Default-Start:  S
 # Default-Stop:   
 # Description:    Generate ssh host keys if they do not exist
 ### END INIT INFO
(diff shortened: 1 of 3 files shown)


mengelmann:
Note to me: Need to check backward compatibility with wheezy - check if keygen supports ed25519, as @andsens noted in #431. 

Wheezy benefits from Long Term Support (LTS) until the end of May 2018.

mengelmann:
The more you get into it, the more complicated it becomes:
- wheezy: DSA, RSA, ECDSA (and with backports also ED25519, see below)
- jessie: DSA, RSA, ECDSA, ED25519
- stretch: RSA, ECDSA, ED25519

oppenssh-server from wheezy-backports supports ED25519 keys, so 'generate-ssh-hostkeys' script would have to check at runtime if ED25519 is supported or not. My suggestion is to leave wheezy as is right now in bootstrap-vz (supports RSA, DSA and ECDSA) and just support ED25519 for jessie and newer. 

I added changes in the code to reflect the above. 
PR is also a workaround for #432 (changes LSB header in sysv init script for stretch, see mentioned issue for details).

Please review.

andsens:
Oh wow. Yeah, that's far from simple. Good job on getting through this! Merging...

PR #428, opened 2018-01-08
Change location of expand-root systemd unit from /lib/... to /etc/...

mengelmann:
Change location of expand-root systemd unit from ```/lib/systemd/system``` to ```/etc/systemd/system```.

System administrators should put customized unit files in ```/etc/systemd/system``` and ```/lib/systemd/system``` should contain units provided by installed packages (deb).

https://www.freedesktop.org/software/systemd/man/systemd.unit.html and https://wiki.debian.org/systemd


andsens:
LGTM, merging. Thanks @mengelmann 👍

PR #430, opened 2018-01-11
Cleaning machine ID

### LINKED ISSUE #427: /etc/machine-id is not unset
```/etc/machine-id``` is not unset, so each VM created from the image has the same, not unique machine ID. 

Valid solution is removing content (not the whole file) of ```/etc/machine-id```. I think that it is safe to also clear ```/var/lib/dbus/machine-id``` (contains D-Bus machine ID) if it exists (package ```dbus``` is installed). 

From systemd documentation https://www.freedesktop.org/software/systemd/man/machine-id.html

> The ```/etc/machine-id``` file contains the unique machine ID of the local system that is set during installation or boot.

> For operating system images which are created once and used on multiple machines, for example for containers or in the cloud, ```/etc/machine-id``` should be an empty file in the generic file system image. An ID will be generated during boot and saved to this file if possible. Having an empty file in place is useful because it allows a temporary file to be bind-mounted over the real file, in case the image is used read-only.

### PR DESCRIPTION
**Pull request by mengelmann**
Cleaning machine ID. Fixes #427

### DIFF
--- a/bootstrapvz/common/tasks/cleanup.py
+++ b/bootstrapvz/common/tasks/cleanup.py
@@ -30,3 +30,20 @@ class CleanTMP(Task):
         log = os.path.join(info.root, 'var/log/')
         os.remove(os.path.join(log, 'bootstrap.log'))
         os.remove(os.path.join(log, 'dpkg.log'))
+
+
+class ClearMachineId(Task):
+    description = 'Clearing the Machine ID'
+    phase = phases.system_cleaning
+
+    @classmethod
+    def run(cls, info):
+        import logging
+        log = logging.getLogger(__name__)
+        for machineid_file in [os.path.join(info.root, f) for f in ['etc/machine-id', 'var/lib/dbus/machine-id']]:
+            if os.path.isfile(machineid_file):
+                log.debug(machineid_file + ' found, clearing')
+                with open(machineid_file, 'w'):
+                    pass
+            else:
+                log.debug(machineid_file + ' not found, not clearing')
--- a/bootstrapvz/common/task_groups.py
+++ b/bootstrapvz/common/task_groups.py
@@ -35,7 +35,7 @@ def get_standard_groups(manifest):
     group.extend(security_group)
     group.extend(get_locale_group(manifest))
     group.extend(get_bootloader_group(manifest))
-    group.extend(cleanup_group)
+    group.extend(get_cleanup_group(manifest))
     return group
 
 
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: bootstrapvz/common/tasks/cleanup.py [hunk 1]
```diff
@@ -30,3 +30,15 @@ def run(cls, info):
         log = os.path.join(info.root, 'var/log/')
         os.remove(os.path.join(log, 'bootstrap.log'))
         os.remove(os.path.join(log, 'dpkg.log'))
+
+
+class ClearMACHINEID(Task):
```
andsens: Heh, no need for all uppercase. It's just that MOTD and TMP are actually acronyms. I suggest `ClearMachineId`.

bootstrapvz/common/tasks/cleanup.py
```diff
@@ -30,3 +30,15 @@ def run(cls, info):
...
+    description = 'Clearing the Machine ID'
+    phase = phases.system_cleaning
+
+    @classmethod
+    def run(cls, info):
+        for machineid_file in [os.path.join(info.root, f) for f in ['etc/machine-id', 'var/lib/dbus/machine-id']]:
+            if os.path.isfile(machineid_file):
```
andsens: Check out the "[Roll complexity into which tasks are added to the tasklist](http://bootstrap-vz.readthedocs.io/en/master/developers/contributing.html#roll-complexity-into-which-tasks-are-added-to-the-tasklist)" chapter in the docs. If the file does not exist to begin with the task shouldn't be added to the list during resolution, otherwise you cannot rely on the tasklog to tell you what work was done to create an image.
To do that, I suggest changing `cleanup_group` into a function that checks the release that is being bootstrapped for and conditionally adds the task.

bootstrapvz/common/tasks/cleanup.py
mengelmann: OK

bootstrapvz/common/tasks/cleanup.py
mengelmann: You're right. I was thinking about it and here are some thoughts:
* ```/etc/machine-id``` may not exist if someone installs sysvinit instead of systemd regardless of Debian release.
* ```/var/lib/dbus/machine-id``` is a file from dbus package (the dbus binary creates it, not package or dpkg scripts). dbus package is not installed by default, but someone may install it via ```include_packages```, and IMO it should be cleared even in this case. 

So both files may or may not exist regardless of the release.  Just the probability is different (like 0% for wheezy and 98% for stretch). 

I don't know how to proceed in that case. WDYT @andsens ?

andsens: OK, I would only add it for stretch then and add a debug message to the task if the file does not exist, that's the best you can do really.

mengelmann:
@andsens, I added changes you asked for. Looks better now :)

andsens:
Good stuff. Merging...

PR #431, opened 2018-01-11
Refactorized generation of OpenSSH host keys with a new systemd unit

mengelmann:
Changes:
- New systemd unit unit called "ssh-generate-hostkeys.service" to keep it near "ssh.service".
- Moved content of the old sysvinit script to /usr/local/sbin/ssh-generate-hostkeys (I find sbin better than bin for this type of script).
- I kept old sysvinit script for backward compatibility.  I didn't rewrite the script to "proper" init script with a start/stop/restart arguments as the old one was just a simple bash script.
- Small refactorization of ssh-generate-hostkeys.


andsens:
This looks great! Wow. I'm tempted to merge as-is. What about backwards compat with wheezy though? I don't think we have systemd there, and AFAIK the keygen doesn't support ed25519.

mengelmann:
Wheezy.. right, I forgot. Will you accept the same solution as it is for squeeze right now (conditional with a different script backed into the image)?

Wheezy benefits from Long Term Support (LTS) until the end of May 2018 https://www.debian.org/releases/wheezy/index.en.html

andsens:
> Will you accept the same solution as it is for squeeze right now (conditional with a different script backed into the image)?

Yup, definitely.

> Wheezy benefits from Long Term Support (LTS) until the end of May 2018

Exactly. At that point we can drop support as well.

mengelmann:
Work in Progress. Needs changes due to #426.

mengelmann:
PR updated.

Changes:
* New systemd unit named ```ssh-generate-hostkeys.service``` (name change to keep it near ```ssh.service```). It is deployed only for releases >= stretch.
* SSH host keys generation logic moved to ```/usr/local/sbin/ssh-generate-hostkeys``` with small refactorisation (I find ```sbin``` better than ```bin``` for this type of script).
* New "full" sysvinit script ```/etc/init.d/ssh-generate-hostkeys``` just in case someone does not like systemd.
* Script named ```/etc/init.d/generate-ssh-keys``` is kept for backward compatibility.

I'm not sure if keeping script named ```/etc/init.d/generate-ssh-keys``` is important. 

If you find this PR too intrusive, please let me know :)

andsens:
Nice job!  
> I'm not sure if keeping script named /etc/init.d/generate-ssh-keys is important.

Nah, I would be surprised if anyone interacted with it manually. It's a one-off thing and the rename makes a lot of sense, go for it. Backwards compatibility should be kept for the manifest itself, but we can't start making guarantees for the resulting images, that'd be way too restrictive.

mengelmann:
OK, I removed ```generate-ssh-hostkeys```.

andsens:
Looks great! Merging...

PR #434, opened 2018-01-17
tmpfs_workspace: Plugin for having workspace on tmpfs for speed

liori:
A proof-of-concept with no documentation yet, for discussion.

I noticed that a full image build for one of my images takes about an hour on my home server with no SSD, but more than enough RAM. Tried mounting `tmpfs` over the workspace directory, and I started getting builds in 3 minutes. The idea sounds generic enough for other people to also benefit, so I wrote a plugin.

I should probably split the `MountTmpfs` task into two, to allow proper clean-up if `makedirs` succeeds, but `mount` fails.

I'd prefer not having an additional directory for `CreateWorkspace` to run, but I didn't find any good way to do so. I cannot replace/delete `CreateWorkspace` from a plugin. My first thought was to make `MountTmpfs` run after `CreateWorkspace` and simply mount `tmpfs` over the directory created by `CreateWorkspace`, but the dependency structure is such that I couldn't make `MountTmpfs` a successor of `CreateWorkspace`, nor `UnmountTmpfs` a predecessor of `DeleteWorkspace`, without explicitly listing all tasks from the `preparation`/`cleanup` phases that still perform some I/O inside the workspace directory. And listing them explicitly seems error-prone. So I decided to let `CreateWorkspace` do its thing on some directory that's already on tmpfs. Is there a better way?

If the current approach is fine, I'll write some documentation, split `MountTmpfs` and submit a more polished commit.

### DIFF
--- a/bootstrapvz/plugins/tmpfs_workspace/README.rst
+++ b/bootstrapvz/plugins/tmpfs_workspace/README.rst
@@ -1,0 +1,18 @@
+tmpfs workspace
+---------------
+
+The ``tmpfs workspace`` plugin mounts a tmpfs filesystem for the
+workspace temporary files. This is useful when the workspace directory
+is placed on a slow medium (e.g. a hard disk drive), the build process
+performs lots of local I/O (e.g. building a vagrant box), and there is
+enough RAM to store data necessary for the build process. For example,
+the ``stretch-vagrant.yml`` manifest file from the examples directory
+takes 33 minutes to build on the plugin author's home server. Using
+this plugin reduces this time to 3 minutes at the cost of 1.2GB of
+additional RAM usage.
+
+Settings
+~~~~~~~~
+
+This plugin has no settings. To enable it add ``"tmpfs_workspace":{}``
+to the plugin section of the manifest.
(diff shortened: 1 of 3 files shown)


andsens:
Whoa, that's an incredible performance boost :-)

Why don't you just replace the CreateWorkspace task and create the folder yourself like [the prebootstrapped plugin does](https://github.com/andsens/bootstrap-vz/blob/d725d3f3bf033a7cc36711442ef9b5500c03a69b/bootstrapvz/plugins/prebootstrapped/__init__.py) (name it `CreateTmpFsWorkspace` or something)? Seems to me that is the easiest option.  
Task dependencies will still work, because they are calculated on all tasks that exist, rather than the ones in the tasklist.

liori:
Thank you for the suggestion of using `taskset.discard`. I think I misunderstood it when I was looking for a similar mechanism. The current version of the commit is using it now. Also, I split each task into two smaller ones to have nicer failure recovery and added some minimal documentation.

andsens:
Great work. Good job on the docs, it should be plenty.  
> Also, I split each task into two smaller ones to have nicer failure recovery

Yup, additionally this makes it possible to mess with the tasks in a more fine-grained manner from other code in the future.

PR #435, opened 2018-01-18
Minor changes to GRUB configuration regarding console and serial console

mengelmann:
Minor changes to GRUB configuration regarding virtual and serial consoles:
* Corrected GRUB key/variable.
* Print kernel log messages that precede the initialisation of the traditional console to a serial console.
* Add console settings to both - default and recovery GRUB options (```GRUB_CMDLINE_LINUX``` instead of ```GRUB_CMDLINE_LINUX_DEFAULT```).

Above changes do not influence GRUB configuration modifications made by providers (Azure, GCE, EC2, VirtualBox).

GRUB documentation https://www.gnu.org/software/grub/manual/grub/grub.html#Simple-configuration 

Kernel documentation https://github.com/torvalds/linux/blob/master/Documentation/admin-guide/serial-console.rst

### DIFF
--- a/bootstrapvz/common/tasks/grub.py
+++ b/bootstrapvz/common/tasks/grub.py
@@ -238,7 +238,7 @@ class SetGrubTerminalToConsole(Task):
     @classmethod
     def run(cls, info):
         # See issue #245 for more details
-        info.grub_config['TERMINAL'] = 'console'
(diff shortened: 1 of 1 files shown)


andsens:
Thanks. Good call on `earlyprintk`, that makes sense. Merging...

PR #436, opened 2018-01-19
Add to the KVM option to set output of the init to a virtual or serial console

mengelmann:
PR adds to the KVM provider option to set output of the init process to a virtual or serial console. 

Output (```stdout```, ```stderr```) of the init process (e.g. systemd) cannot be directed concurrently to both virtual and serial console - only the last device specified as ```console=XXX``` will show startup messages and act as a console in single-user mode. **So order matters.**

The output of kernel messages generated by ```printk()``` and seen by ```dmesg``` always goes to the serial console and the virtual one if the option in a manifest file is set.

Serial console stays as default for backward compatibility.

I added this only to KVM because the virtual console does not make sense for:
* cloud providers (Azure, EC2, GCE, Oracle) - no visual output
* Docker - no kernel boot process
* VirtualBox - already adds ```tty0```

BTW, regarding VirtualBox - I'm not sure if the order of tasks in phase is predictable and if ```successors = [grub.WriteGrubConfig]``` is enough to be sure that this task adds ```tty0``` as the last console (```
predecessors = [grub.SetGrubConsolOutputDeviceToSerial]``` may be necessary?) https://github.com/andsens/bootstrap-vz/blob/f00c63eaeb7cd06c28f56e2b228a979dade9a873/bootstrapvz/providers/virtualbox/tasks/boot.py#L9


andsens:
Nice. LGTM! Merging.
> BTW, regarding VirtualBox - I'm not sure if the order of tasks in phase is predictable and if successors = [grub.WriteGrubConfig] is enough to be sure that this task adds tty0 as the last console (predecessors = [grub.SetGrubConsolOutputDeviceToSerial] may be necessary?) 

You're right, it should definitely be a predecessor. Although it would be even better if the two lines in `grub.SetGrubConsolOutputDeviceToSerial` were split up into two tasks, so that you can remove the one that sets `console=ttyS0` when resolving tasks.

PR #437, opened 2018-01-20
Add yamllint linter to check manifests files

mengelmann:
Add http://yamllint.readthedocs.io/ linter to check manifests files. 

I added yamllint environment to tox right after invoking flake8.

One nonadherence from default yamllint rules (```.yamllint```) is line length - changed from 80 to 160 characters. It's because lines that violate this rule are URLs and release names and I don't see a point in making it more complicated than it should be.

### DIFF
--- a/.yamllint
+++ b/.yamllint
@@ -1,0 +1,6 @@
+---
+extends: default
+
+rules:
+  line-length:
+    max: 160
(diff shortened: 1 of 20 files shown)


andsens:
Aw yeah. I love this. Everything that makes the code more standards compliant gets a big thumbs up from me 👍 👍 👍 :-)  
Merging...

PR #438, opened 2018-01-20
Force order of tasks that add console=xxx to kernel command line (VirtualBox)

mengelmann:
Force order of tasks that add console=xxx to kernel command line (VirtualBox), as we discussed in https://github.com/andsens/bootstrap-vz/pull/436#issuecomment-359170266.


andsens:
👍

PR #439, opened 2018-01-20
Add comment-header to files deployed by bootstrap-vz

mengelmann:
Adds comment-header to files deployed by bootstrap-vz, so it is easier to check where does the file come from (distribution, bootstrap-vz, manually).

I also added a header to GRUB config file as it is format is different than the one that comes from distribution. 

I did not add headers to apt sources, preferences and configuration files because bootstrap-vz provides defaults only for apt sources and these are standard and simple. I would like not to add headers to files created via a manifest file (user knows what he asked to add).


andsens:
Wow, that's a lot of work. Thank you @mengelmann.  
One thing, instead of
```
# This is part of bootstrap-vz: http://github.com/andsens/bootstrap-vz
# See https://github.com/andsens/bootstrap-vz/blob/master/LICENSE for
# legal notices and disclaimers.
```

wouldn't this be better?
```
# This file was created by bootstrap-vz.
# See https://github.com/andsens/bootstrap-vz/blob/master/LICENSE for
# legal notices and disclaimers.
```

We don't need to specify the URL twice, just referencing the license seems fine.
Also, the wording is slightly improved.  

Running
```
find . -type f -exec sed -i s%# This is part of bootstrap-vz: http://github.com/andsens/bootstrap-vz%# This file was created by bootstrap-vz.%
```
should fix it I believe.

mengelmann:
You're absolutely right :) Please see the new commit.

andsens:
Awesome. Merged :-)

PR #440, opened 2018-01-20
Change location of expand-root script and drop .sh extension

mengelmann:
Changes the location of the expand-root script from ```/usr/bin/``` to ```/usr/local/sbin``` as it's more suitable for scripts not installed from deb packages and drops ```.sh``` extension.

Related to #428.


andsens:
Makes sense. Merging...

PR #441, opened 2018-01-21
qemu-img also supports shrinking raw and qcow2 formats

mengelmann:
qemu-img also supports raw and qcow2 formats, so let's allow shrinking them too.

```
$ qemu-img --help | grep ^Supported
Supported formats: blkdebug blkreplay blkverify bochs cloop dmg file ftp ftps gluster host_cdrom host_device http https iscsi iser luks nbd nfs null-aio null-co parallels qcow qcow2 qed quorum raw rbd replication sheepdog ssh vdi vhdx vmdk vpc vvfat
```


andsens:
LGTM, merging...

PR #446, opened 2018-01-25
Tag ebs volumes when using ec2 provider

vmlintu-nosto:
Add tags also to the EBS volume when it is created. The tags can be used to set more specific IAM policies that limit operations on the volumes.

The same tags are already set for ebs snapshots and AMIs.

### DIFF
--- a/bootstrapvz/providers/ec2/README.rst
+++ b/bootstrapvz/providers/ec2/README.rst
@@ -169,6 +169,31 @@ Example:
       region: us-west-1
 
 
+Tags
+~~~~
+
+EBS volumes, snapshots and AMIs are tagged using AWS resource tags
+with the tag names and values defined in the manifest. Tags can be used to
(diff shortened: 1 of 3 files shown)


andsens:
Nice. But I can't merge this without some docs explaining the feature.

vmlintu-nosto:
I couldn't find documentation describing the current tagging functionality for snapshots and AMIs, so I added a piece describing all of them in the ec2 provider README. The tag values seem to allow placeholders, but I didn't yet find what is supported.

vmlintu-nosto:
The tag data creation code itself follows the example of the snapshot tagging code:

https://github.com/andsens/bootstrap-vz/blob/master/bootstrapvz/providers/ec2/tasks/ebs.py#L32

vmlintu-nosto:
I changed now tags to be a named parameter and it defaults to an empty array.

andsens:
Great! Merging.  
Thank you @vmlintu-nosto!

PR #447, opened 2018-01-26
add ability to install tasksel tasks

CMeza99:
```yml
package:
  tasksel:
    - ssh-server
```

need to add docs


andsens:
60dbbee731fd3b07dcdbad22a203c8fef858375c introduced the option to install the tasksel standard package. The reason the other tasksel options were not introduced is because you can install them via apt, i.e. `apt-get install ssh-server`. tbh I don't see the need to introduce this feature, one should just add the task to the package.

CMeza99:
fair point

PR #448, opened 2018-02-01
select version for amzn-driver

CMeza99:
Add ability to declare version of the Amazon driver for EC2 instances. Currently, bootstrap-vz is pulling from master, but the version is incorrectly labeled at `1.0.0`.

I need to add docs, but I wanted feedback first.

### DIFF
--- a/bootstrapvz/providers/ec2/README.rst
+++ b/bootstrapvz/providers/ec2/README.rst
@@ -124,6 +124,28 @@ Example:
       virtualization: hvm
       enhanced_networking: simple
 
+Amazon Drivers
+~~~~~~~~~~~~~~
+
+Define the version for the Amazon Elastic Network
+Adapter (ENA) driver.
+Read more about this on the `Amazon Drivers git repo`__.
+
+__ https://github.com/amzn/amzn-drivers
+
+-  ``amzn-driver-version``: Default: master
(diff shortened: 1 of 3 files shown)


andsens:
Looks great. Instead of cloning with git you could just download the tar file, it's 1 dependency less. The URLs are pretty simple to construct: The template would be `https://github.com/amzn/amzn-drivers/archive/ena_linux_{version}.tar.gz`

CMeza99:
Yes I agree, less dependencies is better. But that would make this a required field, and break existing templates, or require hardcoding a default version. As a compromise I would like to give the option to specify `git` if someone want to use `HEAD` and leave that as the default, which is what is happening now in the code.

andsens:
So just make `version=master` the default :-)

CMeza99:
yes :man_facepalming:

CMeza99:
I just notice the edit to you previous comment. I followed the same pattern for the ixgbevf driver, for consistency in the code (so I do not need `--strip-components 1 `).

Also, I did not included the `-v`. I think that would add more noise to the debug logs than anything useful. If you disagree, I can added it.

Let me know, if its good and I will add some docs. thx

andsens:
Ah yes, makes sense. Nice job :-)

Merging...

Issue #280, opened 2016-02-07
Remove Squeeze support


myhro:
As yesterday (2016/02/06), `squeeze` [is not supported anymore][1], as it has completed five years since its launch and LTS period has ended.

Any tasks/workarounds/etc. created to specifically support this release should be removed from the code base and documentation.

[1]: https://www.debian.org/releases/squeeze/

andsens:
> Any tasks/workarounds/etc. created to specifically support this release should be removed from the code base and documentation.

Agreed. Though we should tag the commit before that, just so people can get the latest version that bootstraps squeeze.

myhro:
Anders,

Is it okay if I tag 26d09d3 as `0.9.9-squeeze`? Because 790f2d6 already removes `squeeze` mentions from the documentation about supported builds.

andsens:
Yup, that's fine. Go right ahead.

myhro:
Done: https://github.com/andsens/bootstrap-vz/tree/v0.9.9-squeeze

P.s.: there was a [`before-wip-integration-testing-rebase`][1] tag which was pushed together with the intended one. Should we delete it?

[1]: https://github.com/andsens/bootstrap-vz/tree/before-wip-integration-testing-rebase

myhro:
There's an [ongoing discussion][1] about when Squeeze LTS support will really end. It maybe done at the [end of the month][2], but the kernel from 2.6 series had [its support finished already][3].

[1]: https://lists.debian.org/debian-lts/2016/02/msg00036.html
[2]: https://lists.debian.org/debian-lts/2016/02/msg00044.html
[3]: https://lists.debian.org/debian-lts/2016/02/msg00045.html

PR #449, opened 2018-02-07
Remove squeeze support

CMeza99:
Removing EOL Squeeze support. Address issue #280.


bootstrapvz/base/__init__.py
```diff
@@ -15,22 +15,13 @@ def validate_manifest(data, validator, error):
     from bootstrapvz.common.tools import rel_path
     validator(data, rel_path(__file__, 'manifest-schema.yml'))
 
-    from bootstrapvz.common.releases import get_release, squeeze
-    release = get_release(data['system']['release'])
-
-    if release < squeeze:
```
andsens: Instead of removing this, I'd change it to wheezy.

CMeza99: yup. you're right.

andsens:
mmh, feels good. I love removing legacy code :-)

PR #451, opened 2018-02-08
Add apt_repo plugin to add additional apt repos during image build

vmlintu-nosto:
apt_repo plugin supports adding additional apt repositories that can be used to install packages during image build. To support S3 backed apt repos, apt-transport-s3 and ca-certificate packages are
installed during debootstrap.

### DIFF
--- a/bootstrapvz/plugins/apt_repo/README.rst
+++ b/bootstrapvz/plugins/apt_repo/README.rst
@@ -1,0 +1,38 @@
+Add additional apt repos
+------------------------
+
+This plugin adds support for adding additional apt repositories. It imports
+the public repo key from the given URL and adds the sources list files. Multiple
+repositories are supported.
+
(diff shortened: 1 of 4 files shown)


andsens:
I'm fairly certain that all of this can be done already. The [bootstrapper field](http://bootstrap-vz.readthedocs.io/en/master/manifests/index.html#bootstrapper) has an additional option of including packages during bootstrapping. Adding repo keys and the likes can then be done via the normal apt repo modification fields.

Feel free to reopen if the above does somehow not solve the problem that your plugin solves :-)

vmlintu-nosto:
It looks like I had misunderstood some of the functionality earlier and this can indeed be done without my plugin. Thanks for pointing this out! I think the gpg keychain needs to be stored locally instead of downloading it, but that is no problem in this case.

PR #453, opened 2018-02-09
Compress QCOW2 when shrinking

CMeza99:
Seems like one would want a compressed QCOW2 image if the volume is being shrunk.

### DIFF
--- a/bootstrapvz/plugins/minimize_size/tasks/shrink.py
+++ b/bootstrapvz/plugins/minimize_size/tasks/shrink.py
@@ -70,6 +70,10 @@ class ShrinkVolumeWithQemuImg(Task):
     @classmethod
     def run(cls, info):
         tmp_name = os.path.join(info.workspace, 'shrunk.' + info.volume.extension)
-        log_check_call(
-            ['qemu-img', 'convert', '-O', info.volume.extension, info.volume.image_path, tmp_name])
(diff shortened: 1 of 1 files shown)


bootstrapvz/plugins/minimize_size/tasks/shrink.py
```diff
@@ -70,6 +70,8 @@ class ShrinkVolumeWithQemuImg(Task):
     @classmethod
     def run(cls, info):
         tmp_name = os.path.join(info.workspace, 'shrunk.' + info.volume.extension)
-        log_check_call(
-            ['qemu-img', 'convert', '-O', info.volume.extension, info.volume.image_path, tmp_name])
+        shrink_cmd = ['qemu-img', 'convert', '-O', info.volume.extension, info.volume.image_path, tmp_name]
+        if info.volume.extension == 'qcow2':
+            shrink_cmd.insert(4, '-c')
```
andsens: I think you need to add some explanation for what this does `insert(4, '-c')` might be a bit... terse :-)

Too bad qemu-img doesn't have long options.

CMeza99:
Yup, that definitely should be clarified. I added some comments, will that suffice?

andsens:
Plenty sufficient :-)  
Merged.

mengelmann:
I know it's merged already, but shouldn't the compression be configurable? There is a performance impact each time when reading from a compressed qcow2 image.

andsens:
I was unaware. @CMeza99?

PR #454, opened 2018-02-09
Fix example docker manifest

CMeza99:
The new line at the end of the description value was breaking "Creating docker image"


andsens:
Nice catch. Thank you!

PR #455, opened 2018-02-10
[RFC] [WIP] Abstract release information out of the code

CMeza99:
I think it would be a better practice to store the release information outside of the code. This is my initial attempt at that. I would like feedback on my approach and if I am going down the correct path.


andsens:
Right on! That's quite a task to take on. Kudos!  
I think your approach is the right one here. Creating yaml files for what the different distros look like is a nice way to handle this and makes it easy for others to understand.  
I am however a little uncertain whether that approach can be applied consistently throughout the project. You might run into scenarios where one distro requires some more logic than a simple yaml key lookup to be configured correctly while the others are plain and simple. I wouldn't call that a dealbreaker though. It's just important to remember to use the coding escape-hatch at the right time so that you don't end up implementing complex decision trees in a yaml file.

PR #458, opened 2018-02-10
Add Docker specific settings

CMeza99:
From https://github.com/moby/moby/blob/e2e5d4bc9da5ba17bb2822909611f8300fdd80f0/contrib/mkimage/debootstrap,
- setting dpkg unsafe-io
- allowing kernel autoremove

### DIFF
--- a/bootstrapvz/plugins/minimize_size/__init__.py
+++ b/bootstrapvz/plugins/minimize_size/__init__.py
@@ -2,7 +2,7 @@ import tasks.mounts
 import tasks.shrink
 import tasks.apt
 import tasks.dpkg
-from bootstrapvz.common.tasks import locale
+from bootstrapvz.common.tasks import dpkg, locale
 
 
 def get_shrink_type(plugins):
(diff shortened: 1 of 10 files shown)


andsens:
Nice job, especially with the references. I wouldn't add tasks from other plugins though (`CreateDpkgCfg`). I think it'd be much better to simply add a proper manifest example that causes that task to be added.

CMeza99:
Yeah, using another plugin task like that felt dirty. So, I added at as a requirement. I also updated the docs and included two new manifest files for Docker.

bootstrapvz/plugins/minimize_size/__init__.py
```diff
@@ -54,8 +54,8 @@ def resolve_tasks(taskset, manifest):
         if apt.get('autoremove_suggests', False):
             taskset.add(tasks.apt.AptAutoremoveSuggests)
     if 'dpkg' in manifest.plugins['minimize_size']:
-        filter_tasks = [tasks.dpkg.CreateDpkgCfg,
-                        tasks.dpkg.InitializeBootstrapFilterList,
+        taskset.add(tasks.dpkg.CreateDpkgCfg)
```
andsens: What's the motivation for this change?

CMeza99: The path for the dpkg config needs to be created so the setting `force-unsafe-io`, so I wanted to make sure the task CreateDpkgCfg was called regardless of is set in the manifest file for` minimize_size: { dpkg: {} }`. But that still seems hacky. So I have change to just creating the dir in the Docker provider.

andsens:
The manifests & docs look great! Instead of using `makedirs` though, you should probably just move `CreatDpkgCfg` to `common/` so that both plugins may add and depend on it.  
After all that's partially why `common/` exists :-)  
The tasklist is a set, so there is no harm in adding a task twice.

CMeza99:
updated :)

andsens:
Perfect! Merging...

PR #460, opened 2018-02-11
Use upstream json_minify

CMeza99:
Change to use upstream  json_minify instead of bootstrap-vz maintaining its own copy.


andsens:
Niiice, I didn't know there was a package.  
Merging...

PR #461, opened 2018-02-11
Ability to create non-compressed QCOW

CMeza99:
Add new shrink type to the minimize_size plugin, `qemu-img-no-compression`, to add the ability to create non-compressed QCOWs


andsens:
You need to fix some indentation issues, so the tests don't fail.

CMeza99:
corrected :)

andsens:
Looks great! merging...

PR #444, opened 2018-01-23
Make the workspace path relative to the manifest file

mengelmann:
Make the workspace path relative to the manifest file (resolves #424).

I had to modify ```minimize_size``` plugin. 

I left individual commits to show the process flow.

PR #465, opened 2018-03-10
Allow relative paths to workspace

liori:
I like having my paths relative.

I've noticed that wherever a plugin wanted to allow relative paths, yet some tool used by that plugin required an absolute path, the path was converted from relative to absolute somewhere close to the place where the absolute path was required.

At first, I tried to do the same. I started patching all places that used the workspace path, and that would fail at runtime with just a relative path. But it turns out there's quite a lot of places that fail; my somewhat minimalistic test case already required 7 changes.

Then I realized it would be better to have the path relative to the manifest file, not just the current directory—just like the ["vbox: Make guest additions path relative to manifest"](https://github.com/andsens/bootstrap-vz/commit/bf7525426b40d37403d84076132219a0bb839a0a) patch. So, I'd essentially need to patch all the places.

So, I decided to just modify the manifest itself during the "parsing" step. I've noticed that the current code carefully avoids modifying the manifest data structure, probably for a good reason… but this just feels the right approach to me here. Though, maybe there is a better way?

### DIFF
--- a/bootstrapvz/base/manifest.py
+++ b/bootstrapvz/base/manifest.py
@@ -5,6 +5,7 @@ invocations should have etc..
 from bootstrapvz.common.exceptions import ManifestError
 from bootstrapvz.common.tools import load_data, rel_path
 import logging
+import os.path
 log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


andsens:
Points for a creative and simple solution, but this is something that I would prefer to be fixed in a more thorough manner. See #444, maybe you can add your experience and get the PR going again? It introduces the concept of a workspace variable separate from the manifest, which represents the same thing but would work regardless of whether you specify the path in the manifest in a relative or absolute manner.

PR #467, opened 2018-03-28
Create device dirs if they do not exist

vitaminmoo:
Some build environments (e.g. Google Cloud Container Builder docker
images) don't have enough of a /dev filesystem to have
/dev/disk/by-uuid/ by default. Make them if they don't exist.


andsens:
Yup, normally we should have checked for specific platforms where this issue is present and try to create the dir only then, but.. seeing how I am also simply creating the links when they don't exist, adding the directory for it is not much of a stretch really :-)

Merging.

PR #468, opened 2018-03-29
Add support for encrypted AMIs

### LINKED ISSUE #405: Creating encrypted AMI snapshots in Amazon EC2
Do you plan to support creating encrypted AMIs? 

http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/AMIEncryption.html

The official Debian images are not encrypted. It's also not possible to create a copy of such an AMI and encrypt it due to some policy applied to those AMIs. So whenever we create an instance from an unencrypted AMI, we have to go through the following steps to encrypt the EBS volume it uses:

1) Shutdown the new EC2 instance
2) Create a snapshot of the EBS volume and delete the unencrypted volume
3) Create an encrypted copy of the snapshot in the same availability zone
4) Create a new EBS volume based on the encrypted snapshot
5) Attach the encrypted EBS volume as a root device to the instance
6) Launch the instance again

I had a quick look at the manifest:

https://github.com/andsens/bootstrap-vz/blob/master/bootstrapvz/providers/ec2/manifest-schema.yml

but I don't see any settings related to encryption here.

vmlintu-nosto:
This adds support for using encrypted EBS volume with EC2 provider. The resulting snapshot and AMI are encrypted and they can be used to launch EC2 instances with encrypted root volume. Custom KMS key can be used by specifying kms_key_id or otherwise the default kms key is used.

More information about AMI encryption:

https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/AMIEncryption.html

Closes #405 

This pull request conflicts with https://github.com/andsens/bootstrap-vz/pull/446 that I can update if this is merged first.

### DIFF
--- a/bootstrapvz/providers/ec2/ebsvolume.py
+++ b/bootstrapvz/providers/ec2/ebsvolume.py
@@ -4,18 +4,26 @@ from bootstrapvz.base.fs.exceptions import VolumeError
 
 class EBSVolume(Volume):
 
-    def create(self, conn, zone, tags=[]):
-        self.fsm.create(connection=conn, zone=zone, tags=tags)
+    def create(self, conn, zone, tags=[], encrypted=False, kms_key_id=None):
+        self.fsm.create(connection=conn, zone=zone, tags=tags, encrypted=encrypted, kms_key_id=kms_key_id)
 
     def _before_create(self, e):
         self.conn = e.connection
         zone = e.zone
         tags = e.tags
         size = self.size.bytes.get_qty_in('GiB')
-        self.volume = self.conn.create_volume(Size=size,
-                                              AvailabilityZone=zone,
-                                              VolumeType='gp2',
-                                              TagSpecifications=[{'ResourceType': 'volume', 'Tags': tags}])
+
+        params = dict(Size=size,
+                      AvailabilityZone=zone,
+                      VolumeType='gp2',
+                      TagSpecifications=[{'ResourceType': 'volume', 'Tags': tags}],
+                      Encrypted=e.encrypted)
+
+        if e.encrypted and e.kms_key_id:
+            params['KmsKeyId'] = e.kms_key_id
+
+        self.volume = self.conn.create_volume(**params)
+
         self.vol_id = self.volume['VolumeId']
         waiter = self.conn.get_waiter('volume_available')
         waiter.wait(VolumeIds=[self.vol_id],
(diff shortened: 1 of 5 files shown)


bootstrapvz/providers/ec2/tasks/ebs.py
```diff
@@ -8,7 +8,12 @@ class Create(Task):
 
     @classmethod
     def run(cls, info):
-        info.volume.create(info._ec2['connection'], info._ec2['host']['availabilityZone'])
+        # EBS volumes support encryption. KMS key id is optional and default key
+        # is used when it is not defined.
+        encrypted = info.manifest.data['provider']['encrypted'] if 'encrypted' in info.manifest.data['provider'] else False
```
andsens: Instead of this rather long statement, you should use `get()` with a default, like so:  
```
encrypted = info.manifest.data['provider'].get('encrypted', False)
```

bootstrapvz/providers/ec2/ebsvolume.py
```diff
@@ -4,16 +4,24 @@
 
 class EBSVolume(Volume):
 
-    def create(self, conn, zone):
-        self.fsm.create(connection=conn, zone=zone)
+    def create(self, conn, zone, encrypted, kms_key_id):
```
andsens: You have defaults for the manifest of course, but you should probably add some defaults here as well in case somebody uses the code in a plugin to create a secondary ebs volume, for whatever reason.  
This would also make them named parameters, which would work better together with the EBS tag PR.

vmlintu-nosto:
I updated the PR to use named parameters and defaults with get().

bootstrapvz/providers/ec2/tasks/ebs.py
```diff
@@ -16,7 +16,12 @@ def run(cls, info):
             formatted_tags = {k: v.format(**info.manifest_vars) for k, v in raw_tags.items()}
             tags = [{'Key': k, 'Value': v} for k, v in formatted_tags.items()]
 
-        info.volume.create(info._ec2['connection'], info._ec2['host']['availabilityZone'], tags)
+        # EBS volumes support encryption. KMS key id is optional and default key
+        # is used when it is not defined.
+        encrypted = info.manifest.data['provider'].get('encrypted', False)
+        kms_key_id = info.manifest.data['provider'].get('kms_key_id')
```
andsens: This will fail if `kms_key_id` is not set, did you forget a `, None)` ?

vmlintu-nosto: My understanding is that dict.get() returns None for missing objects if nothing else is defined, so I left that out on purpose. Taking out the kms_key_id from the manifest has been working for me without problems. Having the default defined as None would make it more visible, though, so adding it makes sense.

https://docs.python.org/3/library/stdtypes.html

class dict
get(key[, default])
Return the value for key if key is in the dictionary, else default. If default is not given, it defaults to None, so that this method never raises a KeyError.

andsens:
Looks good! Merging...

PR #473, opened 2018-04-27
set permissions for apt gpg pub keys

CMeza99:
Set permission on trusted.gpg file to 644 to address:
`The key(s) in the keyring /etc/apt/trusted.gpg.d/* are ignored as the file is not readable by user '_apt' executing apt-key.`


andsens:
Thanks, weird, this has definitely worked before. Which distribution is this in? Maybe they added a new sanity check.

PR #479, opened 2018-06-13
Fix cloud-init setup broken by pylint changes

vmlintu-nosto:
Commit 701678c9 changes print statements to print functions with end=' '. As the printed strings contain newlines, this causes a space to be written as the first character of the following line causing them to be indented.

An example config that is broken. Here disabling modules indents all the other lines of cloud.cfg except the first one causing the username name setting to fail as it expects indenting to be certain number of spaces.

```
plugins:
  cloud_init:
    metadata_sources: Ec2
    username: admin
    disable_modules:
      - locale
      - mounts
```

This commit removes the spaces by specifying end=''.


andsens:
Good catch. Merging.  
@CMeza99, could you check if your PR may have broken other things in the same way?

PR #482, opened 2018-06-26
python package json_minify needed

citypw:


andsens:
Whoops, yeah. The library was previously included in the code.

PR #485, opened 2018-07-16
Add PNIN as a system manifest parameter.

### LINKED ISSUE #349: DisablePNIN hardwired to debian version > stretch
Hi!

First off, I wanted to thank you for your great work on this tool.  It makes building images of all sorts super easy.  Thanks so much!

I did run into an issue where DisablePNIN is hardcoded to depend on the release version:
```
if manifest.release >= stretch:
            group.append(grub.DisablePNIN)
```
This is probably fine in most cases, but in a case where I am using systemd from jessie-backports, I can't build working ec2 images since that doesn't get called.  Would it be possible to have this hinge on a config option, perhaps with a default based on release?

Thanks!

john-pierce:
Fixes: #349

Adds a new boolean parameter indicating weather to use PNIN or not to
the system section of the manifest.

The default, False enables tasks for extlinux and grub to pass
net.ifnames=0 biosdevnames=0 parameters to the kernel.

Previously for grub PNIN was disabled for stretch and later. This
remains the default but setting system.pnin True will allow the behavior
to be overridden. Extlinux had no prior special behavior,
it now mirrors grub.

### DIFF
--- a/bootstrapvz/base/manifest-schema.yml
+++ b/bootstrapvz/base/manifest-schema.yml
@@ -59,6 +59,7 @@ properties:
         type: string
         pattern: ^\S+$
       locale: {type: string}
+      pnin: {type: boolean}
       release: {type: string}
       timezone: {type: string}
     required:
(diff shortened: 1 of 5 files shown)


john-pierce:
PR description edited to more accurately reflect the implementation, I changed the behavior before submitting without changing the commit message.

Edit: ~~The additional parameters have no effect on earlier
    releases are passed regardless.~~

The parameters are only passed to extlinux if `pnin` is False or unset and release is stretch or greater.

Feel free to edit the commit message while merging.

PR #489, opened 2018-12-10
Adjust chroot environment vars and add --allow-unathuenticated for apt-get

feroda:
I have used the software in ArchLinux and I have experienced some problems during the building of a new image.

They were mainly given by the two arguments in topic.

I added also logging of the env vars in `log_call`.

I have succeeded in creating my custom vagrant box.

### DIFF
--- a/bootstrapvz/common/tasks/apt.py
+++ b/bootstrapvz/common/tasks/apt.py
@@ -213,11 +213,11 @@ class AptUpgrade(Task):
                             'apt-get', 'install',
                                        '--fix-broken',
                                        '--no-install-recommends',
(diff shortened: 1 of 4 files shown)


leegarrett:
I don't understand why you want to allow unauthenticated package installation. That would break the trust chain, and allow MITM attacks on the build. Surely that's not what you intended?

andsens:
Allowing installation of unauthenticated packages is a really bad idea, you should check out how to sign your packages with GPG keys instead.  
Logging the environment with each process invocation will create way too much noise, so I'm not super motivated to merge that either.  

Closing.

PR #494, opened 2019-03-05
Flake8

andrewbogott:
Lots of lint fixes -- this should get the master branch passing Travis again :)

### DIFF
--- a/bootstrapvz/base/fs/partitionmaps/abstract.py
+++ b/bootstrapvz/base/fs/partitionmaps/abstract.py
@@ -69,9 +69,9 @@ class AbstractPartitionMap(FSMProxy):
             # Ask kpartx how the partitions will be mapped before actually attaching them.
             mappings = log_check_call(['kpartx', '-l', volume.device_path])
             import re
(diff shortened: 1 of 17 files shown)


andsens:
Dang! Nice job! Regarding the `\\d` escaping, wouldn't it be easier to just make the string literals raw? As in `r'\d'`?

andrewbogott:
> wouldn't it be easier to just make the string literals raw? As in r'\d'

It might.  I thought I saw some mixed-case examples in there where we had both expanded and non-expanded escapes, but I'll look again and see if that's right.

andrewbogott:
Yeah, there are a couple of ugly cases which mix valid and invalid escapes:

'127.0.0.1\s*{hostname}\n?'

and

'^\troot \(hd0\)$'

I'm going to use double-bashes for those and r'' for the clearer cases.

andrewbogott:
I have another patch ready that addresses those mixed cases... I think they're probably just in error but I don't want to include those changes in this PR since they're (maybe) bugfixes rather than no-ops.

andsens:
Looks great! Merging...

PR #495, opened 2019-03-05
Fix a couple of (presumed) issues with regex escaping.

andrewbogott:
In each of these cases we have an un-escaped \n or \t that
gets passed to a regex.  That means that the regexes actually
contain tabs or newlines, respectively.  After this change
an actual '\t' and '\n' are passed in instead.

I have not tested this!  I am writing this patch based on
the assumption that /surely/ the original authors meant
to pass in \t and \n to the regex and not actual tab
or newline characters.


andsens:
I did. Luckily the python regex engine doesn't mind either way :-)

But this is better of course, merging...

PR #496, opened 2019-03-05
Busterfixes

andrewbogott:
This fixes most of my immediate issues with Buster builds.  It also contains a fix to the puppet plugin, although it currently errors out because puppetlabs hasn't finalized their repo yet.


andsens:
Looking great! Thank you for all the hard work Andrew!

andrewbogott:
Thanks Anders!

